PR #1, opened 2015-08-27
Load correct library depending on OS

cao:


rhelmot:
Could you somehow update this to have a check for `sys.platform == "windows"`?

cao:
I don't have a Windows installation where I can test it, hence I cannot easily add it.

cao:
Windows VM setup and Windows support added :-)

Claripy is untested but should work (same way has been added to Z3, which was tested and working).

zardus:
Thanks!

PR #5, opened 2016-03-19
Fix some imports.

ekse:
I'm getting import errors when generating documentation with Sphinx.

### DIFF
--- a/claripy/backends/backendremote.py
+++ b/claripy/backends/backendremote.py
@@ -1,7 +1,7 @@
 #import time
 
 from ..backend import Backend
-from .backends.z3 import BackendZ3
+from ..backends.backend_z3 import BackendZ3
 try:
     from . import remotetasks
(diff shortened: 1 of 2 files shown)


zardus:
Looks like @andrew did this a few days ago. Sorry I didn't merge it in in time :-(.

On a note regarding this backend, we should convert it to a RemoteFrontend or something. I think that'll be the more modern way of doing it. The performant way of doing it would probably be:

1. CanonicalFrontend, which canonicalizes constraints and evaluated expressions before passing them onto a slave frontend, which would be:
2. CompositeFrontend, which would split constraints into independent sets and use a different frontend for each set. Each of those frontends would be the RemoteFrontend, which would use a remote solver for the constraint solve.

zardus:
Anyways, closing this for now. Again, sorry I didn't get to it in time!

PR #7, opened 2016-03-25
Fixed truism.make_like call paramters in _balance_Reverse

axt:
The truism.make_like function expects a tuple, but it gets called with raw parameters.

### DIFF
--- a/claripy/balancer.py
+++ b/claripy/balancer.py
@@ -240,7 +240,7 @@ class Balancer(object):
     @staticmethod
     def _balance_Reverse(truism):
         if truism.op == '__eq__' or truism.op == '__ne__':
-            return truism.make_like(truism.op, truism.args[0].args[0], truism.args[1].reversed)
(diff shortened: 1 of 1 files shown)


ltfish:
@zardus take a look at it!

ronnychevalier:
If I remember correctly, it was already fixed by @zardus recently

ronnychevalier:
Well master seems to be recent, so I must be wrong, sorry :)

zardus:
I actually did fix it; it's in a merge request on gitlab, but CI issues were holding it up. There's also another fix in that commit, so if you don't mind, @axt, we'll use that one to avoid merge conflicts.

zardus:
Wait, that was before the CI woes, and it was merged (as 4e40b822a4624bf33ba45201d339b04f21c415bd). I wonder why this one isn't showing a conflict...

@axt, are you still experiencing issues with the current version of master?

axt:
It works now, thanks!

PR #12, opened 2016-05-18
Handle an edge case where BVV is passed an empty string (bytes)

iamahuman:
Yes, those edge cases do happen!

### DIFF
--- a/claripy/ast/bv.py
+++ b/claripy/ast/bv.py
@@ -172,12 +172,10 @@ def BVV(value, size=None, **kwargs):
 
     if type(value) is str:
         if size is None:
-            size = 8*len(value)
-            value = int(value.encode('hex'), 16)
-        elif size == len(value)*8:
-            value = int(value.encode('hex'), 16)
-        else:
+            size = len(value)*8
+        elif size != len(value)*8:
             raise ClaripyValueError('string/size mismatch for BVV creation')
+        value = int(value.encode('hex'), 16) if size > 0 else 0
     elif size is None:
         raise ClaripyValueError('BVV() takes either an integer value and a size or a string of bytes')
 


rhelmot:
Can you describe these edge cases? We see a lot of bugs that start with "there's the zero-length string BVV exception being thrown" and end with "there's a bug in the memory model or something that's messing stuff up", so this might obscure some bugs that come up later.

iamahuman:
@rhelmot Actually, I thought passing zero-length string would be perfectly logical despite sounding absurd (you can write zero-length BV into memory which is supposed to do nothing). If it is ever passed an empty string by a bug I think it would probably be an error handling issue (generating empty string rather than throwing an exception). Yes, this certainly would obscure some bugs if it's that case..

ltfish:
> Actually, I thought passing zero-length string would be perfectly logical despite sounding absurd (you can write zero-length BV into memory which is supposed to do nothing).

Thanks for the idea. We discussed about it, and now we have an internal PR that adds support of 0-length BitVectors. Only `Concat` and `Extract` are allowed on 0-length BVs. Arithmetic operations on 0-length BVs are not allowed, because a) the operation does not make sense, and b) we do not mask real mistakes/bugs caused by 0-length BVs.

The internal PR will be merged in after code reviewing, and we'll have to discard this PR than. Sorry :-(

ltfish:
The latest claripy adds a very limited support to 0-length BVVs. Please see this commit: https://github.com/angr/claripy/commit/e17993b774ebcb89422787b560085f5ec5b3a486

I'm closing this PR for now. Thank you again for pointing out this issue.

iamahuman:
Maybe I was aware of edge cases of BVV being made as empty, but not of BVV being used as empty :P

PR #14, opened 2016-05-29
Fix a bug caused by http://bugs.python.org/issue8259

palkeo:
I am shifting with BVV of size 256, and I encountered overflows because of that python2 specificity.

### DIFF
--- a/claripy/bv.py
+++ b/claripy/bv.py
@@ -1,4 +1,3 @@
-import sys
 import functools
 from .errors import ClaripyOperationError, ClaripyTypeError
 from .backend_object import BackendObject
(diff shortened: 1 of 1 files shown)


ltfish:
Umm... how are you even reaching that? I assume you are using a 32-bit OS, and you are trying to shift by a huge amount?

And since this is a BVV (which has a finite length of bits), I think the following logic will work just fine:

```python
if o.signed < self.bits:
	return BVV(self.signed << shift, self.bits)
else:
	return BVV(0, self.bits)
```

palkeo:
Yes, my code was shifting by a enormous amount, triggering that "bug". I figured out later that this enormous shift was caused by a bug on my code…
Feel free to apply your solution, or just to do nothing if you prefer.

BTW, thanks for claripy :) I use it for an analyzer I'm trying to write, and it works very well !

ltfish:
> BTW, thanks for claripy :) I use it for an analyzer I'm trying to write, and it works very well !

@zardus He'll be very happy to see it :-)

zardus:
> BTW, thanks for claripy :) I use it for an analyzer I'm trying to write, and it works very well !

This made my day!

In terms of the fix, we might actually want to apply this. As nice as it is to fail-fast on bugs in the code, Claripy shouldn't blow up with weird arguments such as this. I'll reopen this for now!

zardus:
Okay!

After a surprising amount of internal debate, we've agreed that we need to get this bug fixed. @palkeo, would you mind adapting the fix to what @fish recommended? That way you still get credit for finding and fixing the bug :-)

palkeo:
Here you go :)

zardus:
Thanks!

zardus:
This has been released in Claripy 4.6.6.4.post1 :-)

PR #21, opened 2016-06-28
Update setup.py with the new module.

palkeo:
Without this change, you can't use a claripy downloaded using pip.


ltfish:
Ouch, this is a legit problem.

@zardus: we should add tests for PyPI packages.

PR #22, opened 2016-06-28
Add the __neg__ operation to the concrete backend.

palkeo:
That's because I encountered an error (failed conversion to child node…) because this was missing.

I will probably add more operations if I encounter the same kind of errors :)

### DIFF
--- a/claripy/backends/backend_concrete.py
+++ b/claripy/backends/backend_concrete.py
@@ -21,6 +21,7 @@ class BackendConcrete(Backend):
         self._op_raw['__or__'] = self._op_or
         self._op_raw['__xor__'] = self._op_xor
         self._op_raw['__and__'] = self._op_and
+        self._op_raw['__neg__'] = self._op_neg
 
     @staticmethod
     def BVV(value, size):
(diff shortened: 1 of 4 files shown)


ltfish:
@zardus this one is on you.

ltfish:
@palkeo just curious, what was the AST that triggers this issue? You can print out the AST by calling `repr(ast)` or `ast._dbg_repr()`.

salls:
Thanks for the fix! Could you make a couple changes to it?

The `__neg__` operation should be put inside claripy/bv.py. We have a `class BVV(BackendObject)` which is used by the concrete backend. If you add the `__neg__` operation to that object it will work without any changes to backend_concrete. The operations in backend_concrete are mostly special case operations. 

Then the tests could go in test_bv.py

Also use `1<<a.size()` instead of `2**`, it's faster for large objects. I guess we use `2**` in some places already, but we should change them...

salls:
One more thing, I think your calculation is wrong. Shouldn't `-claripy.BVV(0, 8) == 0`

palkeo:
Ok, what I was trying to do was bitwise NOT, so __invert__, and not __neg__... Hence my test with -0 == 2**size - 1…
(and it was doing that in a code that was trying to do the bitwise NOT of 0xff, and then using it as an AND mask to zero-out the least significant byte of a variable)

I'll update my PR.

palkeo:
Here you go, I hope it's good this time :)

salls:
Thanks! I cherry-picked the most recent commit and merged it. Github doesn't recognize that the merge happened.

Issue #32, opened 2017-01-25
Add check for unicode string value in BVV


ekilmer:
I was playing around and ran into some trouble while processing a unicode string at [1]. It's not too hard to handle a unicode string here. Obtaining the correct byte length for multi-byte unicode characters should still work with `len` on unicode strings. It should also return the correct hex encoded representation.

However, I haven't checked elsewhere for similar issues. Would the better solution be to create a hex string in my program before passing it in? Ex: '\xc3\xab\xcc\x81a\xc3\xba' instead of 'ë́aú'

I can submit a pull-request if you'd like the change that handles unicode.

Thank you!

[1] https://github.com/angr/claripy/blob/master/claripy/ast/bv.py#L182

zardus:
I think having claripy handle unicode automatically could lead to confusion (for example, with copy-pasted string literals that are expected to be ascii but accidentally include some unicode character). I wouldn't be against a PR trying it out by decoding it to a string (and maybe logging a warning or something?).

ltfish:
A thought for the future: does moving to Python 3 automatically solve this problem for us?

rhelmot:
is this resolved now that #33 is merged or are there lingering design issues?

ekilmer:
I haven't run into any other errors or problems with unicode since it was merged, so yeah I would say this is resolved.

PR #33, opened 2017-01-27
Add check for unicode string value in BVV

ekilmer:
ref: #32

### DIFF
--- a/claripy/ast/bv.py
+++ b/claripy/ast/bv.py
@@ -179,7 +179,7 @@ def BVV(value, size=None, **kwargs):
     :returns:       A BV object representing this value.
     """
 
-    if type(value) is str:
+    if type(value) is str or type(value) is unicode:
         if size is None:
             size = 8*len(value)
             value = int(value.encode('hex'), 16) if value != "" else 0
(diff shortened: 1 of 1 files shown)


ltfish:
Does that pass linting? I would suggest `isinstance()` instead of `type()` as a general rule.

ekilmer:
Ah. I was also wondering that. It does pass linting, but I agree with you on `isinstance()` instead of `type()`. I didn't know if you wanted consistency or not. I can change it if you'd like.

zardus:
IIRC, I originally had `type()` in there either for speed reasons.

ltfish:
Here are the testing results from CPython 2.7.13.

```
$ python -m timeit -c "isinstance('abc', str)"
10000000 loops, best of 3: 0.0683 usec per loop
$ python -m timeit -c "type('abc') is str"
10000000 loops, best of 3: 0.0681 usec per loop
$ python -m timeit -c "isinstance(1, str)"
10000000 loops, best of 3: 0.139 usec per loop
$ python -m timeit -c "type(1) is str"
10000000 loops, best of 3: 0.0679 usec per loop
```

So it's much slower for `isinstance()` to fail a type match on an object. This is probably because `int`s have `Number` as their base class. Let's stick to `type()` then.

Also, now we are calling `type(value)` three types. It's better to assign it to a temporary variable and then reuse that variable later.

PR #36, opened 2017-03-09
_extract_simplifier: Directly generate flattened AST when possible.

ltfish:


zardus:
I'm rerunning colorguard, as it failed on the PR test. Might be an intermittent failure, but colorguard and rex are very good at catching simplification problems.

Can you run the rex and tracer testcases on gitlab?

ltfish:
> Can you run the rex and tracer testcases on gitlab?

Do we have angr CI back in the lab?

ltfish:
The colorguard test case failure was obviously a glitch :-)

ltfish:
Ping @zardus

zardus:
I very much didn't forget :-). I don't want to merge this in before we get the rex testcases running -- @salls is almost done with that, so we should be able to merge with confidence soon.

rhelmot:
I think this one is worth reviving.

PR #37, opened 2017-03-09
Fix: Enable replacement cache - it was never used before.

ltfish:
`key in weakref.WeakKeyDictionary` always returns False, which is why
the replacement cache was never really enabled before. Changing it to
`try ... except KeyError ...` solves this issue.

### DIFF
--- a/claripy/ast/base.py
+++ b/claripy/ast/base.py
@@ -574,9 +574,12 @@ class Base(ana.Storable):
 
             hash_key = self.cache_key
 
-            if hash_key in replacements:
-                r = replacements[hash_key]
-            elif not self.variables.issuperset(variable_set):
+            try:
+                return replacements[hash_key]
+            except KeyError:
+                pass
+
+            if not self.variables.issuperset(variable_set):
                 r = self
             elif leaf_operation is not None and self.op in operations.leaf_operations:
                 r = leaf_operation(self)
@@ -715,6 +718,9 @@ class Base(ana.Storable):
         #   if type(old) is not type(new):
         #       raise ClaripyOperationError('cannot replace type %s ast with type %s ast' % (type(old), type(new)))
         #   old._check_replaceability(new)
+
+        if not self.variables:
+            return self
 
         return self._replace(replacements, variable_set=set())
 
--- a/claripy/frontends/replacement_frontend.py
+++ b/claripy/frontends/replacement_frontend.py
@@ -82,9 +82,10 @@ class ReplacementFrontend(ConstrainedFrontend):
         if not isinstance(old, Base):
             return old
 
-        if old.cache_key in self._replacement_cache:
+        try:
             return self._replacement_cache[old.cache_key]
-        else:
+        except KeyError:
+            # not found in the cache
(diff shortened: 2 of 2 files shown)


zardus:
While checking `x in blah` for weak dictionaries is bad (especially when pypy is used), and I think we should merge this, I am very certain that this *did* work before. Membership checking definitely works:

```In [1]: import weakref

In [2]: d = weakref.WeakKeyDictionary()

In [3]: class A(object): pass

In [4]: a = A()

In [5]: d[a] = 10

In [6]: a in d
Out[6]: True
```

ltfish:
Is it the same between CPython and Pypy?

Why are you still awake?

claripy/ast/base.py
```diff
@@ -716,6 +719,9 @@ def replace_dict(self, replacements):
         #       raise ClaripyOperationError('cannot replace type %s ast with type %s ast' % (type(old), type(new)))
         #   old._check_replaceability(new)
 
+        if not self.variables:
```
zardus: This changes the old behavior -- it used to be that you could replace `1` with `2` or whatnot using the ReplacementSolver, whereas with this change, that will no longer be possible. I'm not against this; just something that we need to keep in mind.

ltfish: We never really support replacing concrete ASTs, do we?

ltfish: I mean, with this check `if not self.variables.issuperset(variable_set):` replacing concrete ASTs are essentially disabled.

salls:
Is there a benefit to having
`if not self.variables: return self`

ltfish:
> Is there a benefit to having `if not self.variables: return self`

In my profiling of a tracer run, we are actively trying to replace `False` many times...

claripy/ast/base.py
zardus: `set().issuperset(set())` returns True, so I think we *did* support replacing concrete ASTs.

ltfish: Oh well. I'm going to remove this line then.

zardus:
Why did you delete the PR? I think `try/except` is the correct way to go, and it makes sense not to support concrete ASTs (like I said, we can add an option to disable that check later)...

ltfish:
I'm going to open a new one that does not change the semantics.

But do you figure out why `key in WeakKeyDictionary` returns `False` sometimes? I have to admit that I don't understand why that's the case.

zardus:
That's a bit of a mystery. My goto answer would be because the GC cleaned the AST out from under you, but that doesn't make sense for a `WeakKeyDictionary`.

ltfish:
Here is my test before (during a ipdb breakpoint):

```

zardus:
One thing just occurred to me: the replacement cache is wiped (most of the time) when new replacements are added. Are you sure that wasn't what was happening?

zardus:
Ah, that wouldn't make sense from your example; nevermind.

I have no idea what's going on :-)

ltfish:
Oh I might know what's going on... never mind. Let me rewrite the commit message.

PR #38, opened 2017-03-09
Fix: Make sure replacement cache is always used.

ltfish:
`key in weakref.WeakKeyDictionary` sometimes returns False, which is why
the replacement cache was sometimes not enabled before. Changing it to
`try ... except KeyError ...` solves this issue.


ltfish:
Ref #37.

ltfish:
To explain: my tests were wrong.

PR #40, opened 2017-03-09
Implement an argument filter for flattening simplifier.

ltfish:
This allows us to simplify expressions like (a ^ a ^ b) to b.

### DIFF
--- a/claripy/operations.py
+++ b/claripy/operations.py
@@ -1,6 +1,9 @@
 import operator
 import itertools
 import collections
+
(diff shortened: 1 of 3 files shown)


ltfish:
For reference: it yields a 1000x speed improvement on a tracer run over CROMU_00051.

Thanks @tiffanyb for reporting the speed issue and creating a minimal test case!

zardus:
rex fails on my local machine, but it also does so with master, so I'm going to merge this.

PR #42, opened 2017-04-01
Implemented filters

zardus:
@ltfish, here it is.

Some notes:

1. Currently, filters are not supported for concrete ASTs. Let me know if this is a problem; due to some other optimizations, this is a bit tricky.
2. Conflicts between filter lists `a.filters` and `b.filters` are currently resolved simply by taking the longer list. This is simplistic, but I think it should work for our initial purposes.

Let me know what you think!


zardus:
Ok, @fish, I don't know enough about VFG to debug why this is causing a failure. My guess is that it's due to subtle changes I made to the multi-arg simplifiers, but they look ok to me.

PR #43, opened 2017-04-19
Replacements dict is now copied on state copy

Lukas-Dresel:
Before it was passed by reference which would cause problems when copying states and modifying the replacements


salls:
I'm not sure about this. I think this will cause overhead that we need to avoid in Tracer. I don't want to copy something every step that is likely to grow every couple steps

PR #45, opened 2017-05-10
Do not recurse AST without any replacements

Lukas-Dresel:


salls:
I don't like checks like `if not replacements:`
I much prefer checks like `if len(replacements) == 0:`

ltfish:
@salls That's against PEP8. We should do `if not something:` instead of `if not len(something) == 0` in the future.

salls:
Hmmm. I think we should use the explicit version to avoid bugs, and not use the PEP8 version. It's not hard to hit strange bugs because we aren't explicit enough.

Issue #49, opened 2017-08-05
ClaripyZeroDivisionError in the cache mixin


palkeo:
There is a bug in the mixin that do caching. Here is how to reproduce it:

```
import claripy
 
num = claripy.BVS('num', 256)
denum = claripy.BVS('denum', 256)
e = claripy.BVS('e', 256)
 
# s = claripy.SolverCacheless()  # With this solver, it succeeds
s = claripy.Solver()  # With this solver, it fails
 
s.add(e == 8)
s.satisfiable()
s.add(claripy.If(denum == 0, 0, num / denum) == e)
s.satisfiable()
```

As you can see, with a SolverCacheless() instance, the script works fine. With a Solver() (or CompositeSolver()), it raises an exception in the cache mixin.

PR #51, opened 2017-08-17
Fix for issue #49: ZeroDivisionError in the cache mixin

palkeo:


tests/test_solver.py
```diff
@@ -466,6 +466,23 @@ def test_unsat_core():
         yield raw_unsat_core, s
 
 
+def test_issue49():
```
ltfish: Can you please rename it to a more meaningful name? I agree "issue49" is clear, but I prefer a short string that indicates what the bug is :-)

zardus:
Thanks! I suspect more such bugs are lurking (specifically, evaluating `x/y` where a model is cached with `y == 0`, but other solutions are possible), but this is a great start.

PR #52, opened 2017-08-17
store string instead of integer for z3 ast hash

rhelmot:
This solves the issue I was complaining about in slack.

For the binary that I'm working on, which I can't share, and is also probably malware, this changes angr's behavior from "crash and burn with a z3 exception after about 2000 blocks" to "run out of memory after about 8000 blocks". I suspect that this binary has logic which detects if it's being emulated and tries to fuck up its emulator, like what we did in cgc.

### DIFF
--- a/claripy/backends/backend_z3.py
+++ b/claripy/backends/backend_z3.py
@@ -324,7 +324,7 @@ class BackendZ3(Backend):
         z3_hash = z3.Z3_get_ast_hash(ctx, ast)
         z3_ast_ref = ast.value # this seems to be the memory address
         z3_sort = z3.Z3_get_sort(ctx, ast).value
-        return hash("%d_%d_%d" % (z3_hash, z3_sort, z3_ast_ref))
+        return "%d_%d_%d" % (z3_hash, z3_sort, z3_ast_ref)
 
     def _abstract_internal(self, ctx, ast, split_on=None):
         h = self._z3_ast_hash(ctx, ast)


zardus:
Hash coliisions.....

ltfish:
Any (negative) performance impact? We'll be doing a lot more string comparisons now.

rhelmot:
That's why I submitted this as a PR instead of just pushing it!

ltfish:
I mean, apparently @zardus merged it. I believe he must have some very good answers :-)

salls:
The string comparisons themselves might be negligible. Increased memory pressure could cause the gc to impact performance

ltfish:
I would say both of them matter. I think what we really need is a better hashing algorithm that generates 64-bit integers.

salls:
I thought python checks the hash before actually doing the actual string comparison

ltfish:
Even if the hashes are stored as integers?

zardus:
Yeah, it does, so the comparison should happen only during collisions. I
think the speed decrease is going to be negligible...

On Aug 17, 2017 11:35 AM, "salls" <notifications@github.com> wrote:

I thought python checks the hash before actually doing the actual string
comparison

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/angr/claripy/pull/52#issuecomment-323126564>, or mute
the thread

PR #53, opened 2017-08-18
Make the code more python3-compatible

palkeo:
It's a long way from being compatible with python3, but it should help.

Clean up print statements, and references to long.
Reorder imports, too.

It contains multiple commits, because I have no idea how to use git properly. But the diff is the right one.

### DIFF
--- a/claripy/__init__.py
+++ b/claripy/__init__.py
@@ -59,7 +59,7 @@ ast._import()
 
 def BV(name, size, explicit_name=None): #pylint:disable=function-redefined
     l.critical("DEPRECATION WARNING: claripy.BV is deprecated and will soon be removed. Please use claripy.BVS, instead.")
-    print "DEPRECATION WARNING: claripy.BV is deprecated and will soon be removed. Please use claripy.BVS, instead."
+    print("DEPRECATION WARNING: claripy.BV is deprecated and will soon be removed. Please use claripy.BVS, instead.")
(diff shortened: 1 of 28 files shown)


zardus:
@pcgrosen or @Lukas-Dresel, could you guys rebase `feat/filters` on top of this (or a squashed version of this) so that we can merge it in?

rhelmot:
Does this cause any performance issues? We use `type` instead of `isinstance` for speed reasons.

ltfish:
> Does this cause any performance issues?

We should provide some test cases just for benchmarking.

rhelmot:
From discussion in slack: the `_get_hashables` method seems to be totally unused, so its performance doesn't really matter!

rhelmot:
Almost there - you just have a bunch of lint regressions: https://travis-ci.org/angr/claripy/jobs/270662171#L1391

We use pylint with the pylintrc file found in the angr-dev repo to ensure that our code quality is tolerable :)

rhelmot:
Oh hm - the vast majority of those lint failures are "superfluous parens", but we need those for python 3... I will disable that message in the pylintrc.

palkeo:
I merged, and fixed some lint errors. Travis passes, AppVeyor fails, but I was always the case if I look at the history

ltfish:
Thanks! Appveyor currently does not work. I believe you've fixed everything!

rhelmot:
Awesome, thank you!

PR #56, opened 2017-11-05
Fix: extra_constraints could be ignored when invoking _max() and _min…

ercoppa:
…() due to early solver pop operation


zardus:
Good catch. It's hard to count the number of bugs we've had in these min and max functions.

PR #57, opened 2017-11-13
ite_excaveted triggered  a bug in normalize_address_type

badnack:
when the address expression was a ite, the added annotation in _normalize_address_type (abstract_memory.py) was removed by the ite_excavated property, causing a bug during CFG construction.


ltfish:
Thanks!

PR #61, opened 2017-12-28
Fix model cache

tyb0807:
@zardus as discussed on slack, please merge this and leave #60 here for a while


ltfish:
Wow, I didn't realize that there is a hardcoded limit!

PR #63, opened 2018-01-31
WIP: fix zerodivision issues when cache is used

zardus:
note to self: I want to add some testcases still, and figure out for reals why it never happened on eval.

### DIFF
--- a/claripy/frontend_mixins/model_cache_mixin.py
+++ b/claripy/frontend_mixins/model_cache_mixin.py
@@ -209,7 +209,10 @@ class ModelCacheMixin(object):
         results = set()
 
         for m in self._get_models(extra_constraints):
-            results.add(m.eval_list(asts))
+            try:
+                results.add(m.eval_list(asts))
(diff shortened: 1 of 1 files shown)


zardus:
With the bizarre Z3 zerodivision behavior (below), I think this is good for now.

```In [13]: import z3, claripy
    ...: z = z3.Solver()
    ...: a = z3.BitVec('a', 32)
    ...: b = z3.BitVecVal(0, 32)
    ...: z.add(a == 1)
    ...: z.check()
    ...: print z.model().eval(a/b)
4294967295```

ltfish:
Looks good to me!

zardus:
Yes it is! IIRC, I wanted to push some testcases to this and figure out a way to import the pyvex bit conditionally (probably easiest to just not import it from `__init__.py`.

The idea was that this fix at least stops angr from excepting out when symbolic divisions by zero explicitly happen, and the pyvex fix makes sure they do not happen during VEX simulation.

PR #69, opened 2018-04-03
Fix bitwise xor simplifier

SweetVishnya:
# Fix bitwise xor simplifier
Filtering out all \_\_xor__ args leads to \_\_xor__ over an empty tuple which is undefined without an initial value, so it is followed by TypeError('reduce() of empty sequence with no initial value').
_flatten_simplifier is now provided with corresponding initial value.


ltfish:
Thank you!

PR #70, opened 2018-04-11
BackendZ3: Store a copy of tactics in TLS.

ltfish:
For some reason, when using PyPy, tactics are never freed in Z3 once they
are created.  Caching tactics per thread will reduce the amount of memory
leaked.

### DIFF
--- a/claripy/backends/backend_z3.py
+++ b/claripy/backends/backend_z3.py
@@ -773,7 +773,8 @@ class BackendZ3(Backend):
 
         #s = expr_raw
         if isinstance(expr_raw, z3.BoolRef):
-            s = self._boolref_tactics(expr_raw).as_expr()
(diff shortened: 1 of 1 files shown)


subwire:
I was told to add my support, so here you go!
This resulted in an issue whereby executing a particularly complicated function would eat up a lot of memory, which was never freed when all relevant objects were discarded, but only _after_ you build a CFG!
These things are supposed to be unrelated events, but it looks like this was triggering the above issue, as CFG apparently makes aggressive use of Z3/Claripy in some cases. 

We still leak a bunch of memory in Z3 using the above scenario, but after this fix, the amount is much, much less, and at least does not super-linearly increase until OOM :)

Once again, @ltfish saves the day!

zardus:
Thank you for the support, @subwire! :-)

PR #71, opened 2018-04-12
BackendZ3: Add memory pressure for new solvers to make PyPy's GC happy.

ltfish:


zardus:
Wow, that's quite clean. Not a perfect solution, but I think it's good enough for most cases. I approve!

ltfish:
Reference: http://cffi.readthedocs.io/en/latest/using.html#memory-pressure-pypy

ltfish:
Apparently it's also a problem for CFFI... I guess Z3 is the only native component in angr that uses so much memory and exposes this issue.

PR #75, opened 2018-04-19
Reverse ite cases

rhelmot:
Adds a reverse operation for for ite_cases and tweaks the And simplifier so it produces sane results. The one concern is, as always, whether the simplifier tweaks are too heavy. I don't think they should be, but I'd like another pair of eyes on it.


ltfish:
Looks good to me, although I don’t know why tests fail though. Will take a look once I arrive at the hotel.

zardus:
I don't think this is too heavy. Looks good to me!

PR #77, opened 2018-05-03
Pop solver during min with extra constraints, makes it match max

### LINKED ISSUE #76: Logic error in claripy z3 min function
Located [here](https://github.com/angr/claripy/blob/master/claripy/backends/backend_z3.py#L687).
The return statement causes `extra_constraints` to not be removed from the z3 solver.

Also, consider using the z3 [optimize](https://z3prover.github.io/api/html/classz3py_1_1_optimize.html) object as it works quite well and will probably out perform your multiple evals.

rhelmot:
Closes #76


zardus:
How did any of this stuff ever work...

PR #79, opened 2018-05-15
Smtlib backends for string solving support

Lukas-Dresel:
We implemented solver backends for string solvers as well as AST support for string constraints. This adds a dependency to claripy on pysmt which is what we used to parse/create smtlib constructs. Most of these changes should not impact behavior when string constraints are not being used. Please give it a thorough read-through to see if that is true. 

The solver backends for the different string solvers will only be enabled if the corresponding solver is installed and available in the current path.

The test cases currently important to pass are `test_backend_smt.py` and `test_backend_smt_cvc4.py`, the other solvers (not the backends) either are missing features or have bugs that make the testcases fail. CVC4 will need to be installed for `test_backend_smt_cvc4.py` to run.


ltfish:
Yan and I are working on a proposal together, so it might take me some time to fully review this PR. I declare my tentative approval here :)

But I still want to read through the whole thing - don't merge it in yet!

ltfish:
Also, a note to whoever the merger will be: Make sure you "Squash and merge" this PR.

Lukas-Dresel:
This has been merged in a different PR.

PR #81, opened 2018-05-23
Explicit handling of BVS for BackendConcrete

drone-rites:
When a `BVS` instance is passed to `BackendConcrete` for conversion, the conversion would fail when arguments of the BVS "op" were converted. In this specific case, the arguments for BVS include a string and None objects.

Rather than implicitly fail when processing a `BVS`, the `BackendConcrete` will now raise an error when handling a `BVS` instance. In order for this to fit in with how handlers for other ops are structured, the `BackendConcrete._convert` method not supports basestring and NoneType instances by simply returning the original value.

*NB:* Please verify providing identity conversion for basestring and NoneType objects is ok. If not, we can look at failing on a BVS "op" before arguments are converted in `BackendConcrete._convert`.

### DIFF
--- a/claripy/backends/backend_concrete.py
+++ b/claripy/backends/backend_concrete.py
@@ -1,6 +1,7 @@
 import logging
 import numbers
 import operator
+import types
 from functools import reduce
 
 l = logging.getLogger("claripy.backends.backend_concrete")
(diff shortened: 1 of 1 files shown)


rhelmot:
@zardus pls confirm

mattrepl:
This probably should be done by filtering ops before attempting to eval the arguments. Just didn’t want to change the current pattern without discussion.

I don’t like the addition of supporting strings and None unless that happens to be useful for other cases too.

PR #82, opened 2018-05-28
Fix unconstrained and args[4] getting out of sync on BVSs

Nilocunger:


ltfish:
Can you please add a comment after the line so that we know `args[4]` is the `undefined` flag? Other than that, it is good. Thank you for finding and fixing this bug!

rhelmot:
I just ran into this bug independently. @Nilocunger you should keep a better to-do list...

rhelmot:
I tested this fix a little more and it turns out is _not_ sufficient - this somehow needs to be done inside backend_z3, since this value is a part of the arguments and thus is part of what gets hashed, and thus must be present before object creation.

Furthermore, there's the even more serious issue that this code is only run on the topmost AST, so if a constraint containing an uninitialized variable is simplified, then the resulting constraint will have the same variable but marked as initialized.

Nilocunger:
Yeah, sorry. I hadn't merged it in yet since I realized it wasn't sufficient and haven't had the time to make an actual fix. Thanks for taking care of it!

PR #85, opened 2018-07-27
Spring cleaning july 2018

rhelmot:
- Replace references to 'I' with either BVV or nothing, depending on context
- Replace references to the list of leaf operations with `operations.leaf_operations`
- Add VSA ops to list of leaf operations
- Remove some dead code from the base constructor

partner to angr/angr#1155

### DIFF
--- a/claripy/ast/base.py
+++ b/claripy/ast/base.py
@@ -97,9 +97,6 @@ class Base(ana.Storable):
         :param variables:       The symbolic variables present in the AST (default: empty set)
         :param symbolic:        A flag saying whether or not the AST is symbolic (default: False)
         :param length:          An integer specifying the length of this AST (default: None)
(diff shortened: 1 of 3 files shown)


claripy/ast/base.py
```diff
@@ -975,7 +966,7 @@ def __getattr__(self, a):
             return self
 
 def simplify(e):
-    if isinstance(e, Base) and e.op == 'I':
+    if type(e) is BV and e.op == 'BVV':
```
zardus: I think this one should check all of leaf_operations. There's not much we can do to simplify an `FPV` or a `BoolV`, (or, I guess sometime soon, a StringV), either. Back when this was checking for `I`, it would have matched any immediate, including booleans and floating points.

PR #90, opened 2018-09-11
Fix thread safety by moving extra_bvs_data to a property

subwire:
Remember folks, the contents of tls must be initialized in each thread!
Fixed extra_bvs_data to use the same pattern as the rest of claripy for dealing with this 

Note that libz3 may still segfault when multithreading; we know about it, but have not managed to track it down through aggressive audits.  
Does not happen in all circumstances, i have one symexec-based analysis ripping apart a 48-thread machine right now without issues.


ltfish:
WOW

zardus:
I agree that it's a hack, but it looks reasonable to me.

subwire:
overriding failures per @rhelmot (we should fix fidget at some point!)

PR #104, opened 2018-10-28
Allow reusing per-thread Z3 solvers. Controlled by envvar REUSE_Z3_SOLVER.

ltfish:
Solves the issue of "memory fragmentation" when too many Z3 solvers are created. Should be enabled by default for symbolic tracing or whenever memory pressure occurs.

### DIFF
--- a/claripy/backends/backend_z3.py
+++ b/claripy/backends/backend_z3.py
@@ -560,7 +560,7 @@ class BackendZ3(Backend):
             _add_memory_pressure(1024 * 1024 * 10)
             if reuse_z3_solver:
                 # Store the Z3 solver to a thread-local storage if the reuse-solver option is enabled
-                Z3ThreadSolvers.solver = s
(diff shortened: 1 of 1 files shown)


claripy/backends/backend_z3.py
```diff
@@ -45,6 +45,16 @@ def _add_memory_pressure(p):
...
 supports_fp = hasattr(z3, 'fpEQ')
 
+#
+# Per-thread Z3 solver
+#
+
+reuse_z3_solver = True if os.environ.get('REUSE_Z3_SOLVER', "False").lower() in {"1", "true", "yes", "y"} else False
+
+
+class Z3ThreadSolvers:
```
zardus: Why the dedicated class here, instead of just sticking it into BackendZ3's tls instance?

ltfish: Done.

PR #106, opened 2018-10-29
Make claripy less recursive

danse-macabre:
This PR is an attempt to prevent claripy from hitting recursion limit.

As a side effect, there is also a performance improvement that can be observed on the large ASTs.
The following performance measurments were done on the AST that has depth 44:

```python
# Recursive version
In[30]: timeit('list(v.recursive_leaf_asts)', number=200, globals=locals())
Out[3]: 6.918710868805647

# Non-recursive version
In[32]: timeit('list(v.leaf_asts())', number=200, globals=locals())
Out[5]: 2.1140575450845063

# Recursive version
In[33]: timeit('list(v.recursive_children_asts)', number=200, globals=locals())
Out[6]: 0.9669999810867012

# Non-recursive version
In[34]: timeit('list(v.children_asts())', number=200, globals=locals())
Out[7]: 0.13944317493587732

# Recursive version
In[35]: timeit(stmt='v.depth', number=2000, globals=locals())
Out[8]: 0.7577604330144823

# Non-recursive version
# Almost zero since the depth is cached at AST init time.
```

### DIFF
--- a/claripy/ast/base.py
+++ b/claripy/ast/base.py
@@ -84,12 +84,26 @@ class BaseMeta(type):
         return self
 
     @staticmethod
-    def _calc_hash(op, args, kw):
+    def _calc_hash(op, args, keywords):
         """
         Calculates the hash of an AST, given the operation, args, and kwargs.
+
+        :param op:          The operation.
+        :param args:        The arguments to the operation.
+        :param keywords:    A dict including the 'symbolic', 'variables', and 'length' items.
+        :returns:           a hash.
+
+        We do it using md5 to avoid hash collisions.
+        (hash(-1) == hash(-2), for example)
         """
-        return hash((op, kw['symbolic'], kw['variables'], kw.get('length', None), kw.get('annotations', None),
-                    tuple((a, a // 2**31, a < 0) if type(a) in (int, float) else hash(a) for a in args)))
+        args_tup = tuple(long(a) if type(a) is int and int is not long else (a if type(a) in (long, float) else hash(a)) for a in args)
+        to_hash = (op, args_tup, keywords['symbolic'], hash(keywords['variables']), str(keywords.get('length', None)), hash(keywords.get('annotations', None)))
+
+        # Why do we use md5 when it's broken? Because speed is more important
+        # than cryptographic integrity here. Then again, look at all those
+        # allocations we're doing here... fast python is painful.
+        hd = hashlib.md5(pickle.dumps(to_hash, -1)).digest()
+        return md5_unpacker.unpack(hd)[0]  # 64 bits
 
     @staticmethod
     def _finalize_args(args, **kwargs):
--- a/claripy/ast/base.py
+++ b/claripy/ast/base.py
@@ -1,9 +1,17 @@
+import hashlib
 import itertools
 import logging
+import numbers
 import os
 import struct
 import weakref
 from collections import OrderedDict, deque
+from past.builtins import long
+
+try:
+    import cPickle as pickle
+except ImportError:
+    import pickle
 
 import ana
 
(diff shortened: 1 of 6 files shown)


rhelmot:
This... looks pretty good. I'll need yan to sign off on it but I like it.

For the lint failure you can probably just disable attribute-defined-outside-init for ast/base.py. See if you can get it to 10/10 lint of course :)

claripy/ast/base.py
```diff
@@ -392,165 +390,187 @@ def remove_annotations(self, remove_sequence):
...
+        prevent floods of text being printed.
 
-        :param max_depth:   The maximum depth to print
-        :return:            A string representing the AST
+        :param max_depth:           The maximum depth to print.
+        :param explicit_length:     Print lengths of BVV arguments.
+        :param details:             An integer value specifying how detailed the output should be:
+                                        0 - print short repr for both operations and BVs,
+                                        1 - print full repr for operations and short for BVs,
+                                        2 - print full repr of both operations and BVs.
```
ltfish: I prefer enums :)

claripy/ast/base.py
```diff
@@ -392,165 +390,187 @@ def remove_annotations(self, remove_sequence):
...
+    @staticmethod
+    def _op_repr(op, args, inner, length, details):
+        """_op_repr
+
+        :param op:
+        :param args:
+        :param inner:
+        :param length:
+        :return:
+        """
```
ltfish: Empty docstrings :)

ltfish: We can remove its docstring if you want -- it's a private method anyway.

ltfish:
Linting failed. Please lint your code ([here](https://travis-ci.org/angr/claripy/jobs/447919108)).

This is crazy... Thanks so much man, the recursiveness of claripy was driving me nuts cuz I triggered it way too often when using VSA in the past. Great stuff!

zardus:
oh no, a metaclass!

Given this specific usecase, it might not be too bad, but I am nervous...

danse-macabre:
@zardus Why not give it a try? I think in this case, where we want to control exactly how the instance of Base is created, the metaclass fits perfectly. It will be good to hear what @rhelmot thinks about it.

Anyway, it's just a little experiment which purpose is to fix the *attribute-defined-outside-init* the hard way. The changes can be reverted any time before the merge.

danse-macabre:
So, the fidget test fails because it defines a BiHead, which is a subclass of claripy.Base with redefined arguments and a custom `__new__` ([here](https://github.com/angr/fidget/blob/73ccda73c95081099b0b5e3acdf8a4b083edf489/fidget/bihead.py#L6)) method, which explicitly calls the `__init__` regardless of whether the instance is already initialized (i.e., returned from cache). Do we want to keep that behavior?

As for the colorguard, I'm not quite sure what is the rex module? Is it a patcherex?

zardus:
I'll let @rhelmot speak for fidget, but rex is our automatic exploit generation tool, which is here: https://github.com/shellphish/rex

danse-macabre:
ok, now the only problem left is that with fidget

rhelmot:
I have forgotten all the nuances around the construction of BiHead. If you really really want to make the metaclass stuff work you should futz with it until tests pass. I don't believe it's meant to clear attributes on cache load, so this puzzles me. Perhaps BiHeads were exempted from the cache through some shenanigans.

I am extremely scared of the metaclass. Every single time we have tried to put a metaclass anywhere into angr for any purpose, it has ended up being a wildly bad idea and we have reverted it. Furthermore, if this is merged and it ends up causing issues down the line, I don't exactly have faith in you to be on-call for fixing them.

If all the tests pass and you add some documentation describing exactly how the metaclass interacts with normal usage, I will strongly consider it.

rhelmot:
Yan and I talked about this a bit offline. We'd like you to pull out the metaclass for this stuff for this PR, add `# pylint: disable=attribute-defined-outside-init`, and then submit the metaclass stuff as a new PR. We think that'll make it much easier to move forward.

zardus:
Yeap, handling the scary metaclass as a separate PR will let us merge this faster, make @ltfish happy, and also reduce conflict risk from #107.

danse-macabre:
Agreed. I was a little hasty at pulling those changes. Actually, I just realized that there is a serious drawback here: the current use of metaclass denies a user to implement a custom arguments order for the subclasses, as can be seen in the BiHead case.

I'll revert the changes and make a separate PR.

By the way, what do you think about the new [_calc_hash()](https://github.com/angr/claripy/pull/106/commits/703b46c27bcab463cb6ed642ffe3ad6159bf6004) routine? If I understand correctly, the original algorithm uses a combination of pickle and md5 to avoid hash collisions on integers, e.g., `hash(-1) == hash(-2)` or `hash(2 ** 64) == hash(1)`. The new algorithm doesn't have this issue and also have better performance:

```
In[59]: timeit('v._orig_calc_hash(v.op, v.args, defaultdict(lambda: None))', number=20000, globals=locals())
Out[23]: 0.27928233705461025

In[60]: timeit('v._new_calc_hash(v.op, v.args, defaultdict(lambda: None))', number=20000, globals=locals())
Out[24]: 0.03938790084794164
```

rhelmot:
on the "improve _calc_hash() performance" commit:

```
In [1]: import claripy

In [2]: claripy.ast.Base._calc_hash('BVV', (-1, 32), {'symbolic': True, 'variables': ()})
Out[2]: -1017771568654643016

In [3]: claripy.ast.Base._calc_hash('BVV', (-2, 32), {'symbolic': True, 'variables': ()})
Out[3]: -1017771568654643016
```

:/

Python's hash is not designed to be collision-resistant. Collision resistance is totally critical to this. Trust us, we've tried everything here...

Additionally, you should not implement structurally_match that way. It is specifically designed to compare each element individually, that's why it's named "structurally_match". We generally use it for debugging claripy failures.

rhelmot:
further entries to the peanut gallery of python hash being nonsense:

```
In [5]: hash(float(-1))
Out[5]: -2

In [6]: hash(float(-2))
Out[6]: -2

In [7]: hash(float('NaN'))
Out[7]: 0

In [8]: hash(float(0))
Out[8]: 0

In [9]: hash(float('inf'))
Out[9]: 314159

In [10]: hash(float(314159))
Out[10]: 314159
```

danse-macabre:
OK, I got it.

rhelmot:
Thanks!!!

PR #107, opened 2018-10-31
fix assertion on bool conversion on ASTs

zardus:
`__nonzero__` is no longer used for `assert, if, and bool()`. `__bool__` is the new king.

This should not be merged until we merge https://github.com/angr/claripy/pull/106, to avoid conflict risk.


ltfish:
Driller failed, which scares me a bit. Just hit rerun and let's see what happens.

rhelmot:
Driller was broken for a day on master. I just fixed it a few hours ago
(forgot to merge a branch)

PR #109, opened 2018-10-31
HybridSolver: use exact solver only if approximate one yields too many solutions

danse-macabre:
The idea is that we try the `approximate_frontend` first. The approximation should be able to solve some simple cases with great precision, as well, as more complex ones, with the help of the replacement solver. However, if the approximation yields too many solutions, we should ask the exact solver (which is typically backed by Z3) to narrow the solutions space. The narrowing should occur if some constraints can't be handled by the approximation backend, but is easily solvable by the exact backend.

This behavior is disabled by default, of course.


zardus:
Nice! With this, we can probably move the various `APPROXIMATE_X` options from simuvex into Claripy, and it looks like it'll be much simpler that way. Toward that end, could you support a `max_approximate_solutions` argument that'll override `n` when checking if the approximation is sufficient? The idea is that, when resolving symbolic memory accesses, we could accept a slightly higher level of imprecision as a trade-off for avoiding solving.

danse-macabre:
Thanks! Actually, at first I did implement the `max_approximate_solutions`, but then I realized that the maximum acceptable number of solutions is somewhat different for different cases and that one single value won't fit well. For example, there are different values for `READ|WRITE_TARGETS_LIMIT` in SimAbstractMemory, and also there could be a different number of max jump targets if we consider different contexts.

Therefore, I decided to trust the user on that matter. If they specify that they want at most `n` solutions, then `n` is probably an acceptable number. If we manage to get less than `n` solutions using the approximate solver - very well!

> The idea is that, when resolving symbolic memory accesses, we could accept a slightly higher level of imprecision as a trade-off for avoiding solving.

I think it will be good to implement `READ|WRITE_TARGETS_LIMIT` as a SimOption.

danse-macabre:
@zardus

> We do already have read and write target limits in the form of the concretization strategies, so that works.

True, but SimAbstractMemory still relies on the hard-coded constants ([here](https://github.com/angr/angr/blob/0840819803be3250c4ce42937f63c1d1c9c1a93b/angr/state_plugins/abstract_memory.py#L17)). Is there any way the concretization strategies could be used with abstract memory?

zardus:
Ah. Well, there shouldn't be any fundamental issues, but @ltfish is the authority on that. I think it should be doable.

ltfish:
> Is there any way the concretization strategies could be used with abstract memory?

Yep. Nothing is really stopping it. I just did not implement it.

danse-macabre:
@zardus Could you please provide an idea on what exactly should be tested here? The only thing that comes to mind, is that we should call `eval` with `exact` set to None, and then call the same method with `exact` set to True, and see whether the results differ.

zardus:
So in the Z3 backend, we have a counter for the number of times Z3 has been invoked (https://github.com/angr/claripy/blob/master/claripy/backends/backend_z3.py#L46). I think doing eval with `approximate_first=True` and making sure that a) the results make sense and b) the solve count hasn't changed would be great.

zardus:
What do you mean by disabled? Like, not called from the `__main__` clause? If so, it's cause nosetests finds and executes all the `test_*` functions automatically, and that clause is just for convenience.

zardus:
Anyways, thanks for the contribution! This is great.

I think vague next steps, eventually, would be to restructure the `APPROXIMATE_*` stuff in angr to use this instead of the hacked-up and non-functional way it works currently, maybe with the exception of guard condition approximation.

PR #112, opened 2018-11-03
BackendZ3: Resolve Z3 AST hash collisions.

ltfish:
Thanks to the recent patch that makes claripy less recursive, we are
probably releasing Z3 objects more often than before. A negative
consequence is that our Z3 AST hash, which uses to heavily rely on memory
addresses of Z3 ASTs, now have a much higher chance to collide. This patch
solves this problem in a painful way: Always check if the cached Z3 object
is the same as the one being abstracted, and do not return the cached one
if they are not equal to avoid AST collisions. It will surely impact the
speed, but it is still better than returning incorrect results.

Thanks @basher for reporting another issue that led us to find this bug.

Update: The old fix introduces way too many equality checks, which greatly slows down simplifications when ASTs are big and deep. Here is a better solution: We convert `_ast_cache` into an LRUCache and store both high-level claripy ASTs and the low-level `z3types.Ast` objects. Most importantly, we increment the Z3 reference counter of the `z3types.Ast` object so it will not be immediately released. This allows us to leverage the internal de-duplication mechanism of Z3 (if there is any), which hopefully will use the same pointer for AST sub-trees if they are the same. Hence we can use pointer values of low-level Z3 ASTs to key our `_ast_cache` dict. When a low-level AST is popped out of the `_ast_cache` dict, we decrement its reference counter inside Z3.

### DIFF
--- a/claripy/backends/backend_z3.py
+++ b/claripy/backends/backend_z3.py
@@ -10,6 +10,8 @@ import weakref
 from functools import reduce
 from decimal import Decimal
 
+from cachetools import LRUCache
+
 from ..errors import ClaripyZ3Error
 
 l = logging.getLogger("claripy.backends.backend_z3")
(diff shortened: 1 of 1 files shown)


zardus:
Do you have perf numbers for this by any chance?

ltfish:
Collecting them.

ltfish:
In the end, keying the AST cache with pointer values is still a pretty bad idea, but I don't have any other solution that's better than this one. I guess we can stay with it for now.

ltfish:
Rex and Patcherex are broken because of recent changes in shellphish-qemu.

Issue #111, opened 2018-11-02
bug: RecursionError: maximum recursion depth exceeded


benben2301:
The test code:
```
import angr
import claripy


def check(binary_path):
    proj = angr.Project(binary_path, use_sim_procedures=True, default_analysis_mode='symbolic_approximating', auto_load_libs=False)
    cfg = proj.analyses.CFGEmulated(keep_state=True, enable_advanced_backward_slicing=True, context_sensitivity_level=4)
    vfg = proj.analyses.VFG(cfg, function_start=proj.entry, context_sensitivity_level=2, interfunction_level=4, remove_options={angr.options.OPTIMIZE_IR})
     

if __name__ == '__main__':
    test_program = './test'
    check(test_program)

```
Error:
```
  File "/home/wgh/.virtualenvs/angr/lib/python3.5/site-packages/claripy-8.18.10.25-py3.5.egg/claripy/ast/base.py", line 887, in ite_excavated
  File "/home/wgh/.virtualenvs/angr/lib/python3.5/site-packages/claripy-8.18.10.25-py3.5.egg/claripy/ast/base.py", line 831, in _excavate_ite
  File "/home/wgh/.virtualenvs/angr/lib/python3.5/site-packages/claripy-8.18.10.25-py3.5.egg/claripy/ast/base.py", line 831, in <listcomp>
  File "/home/wgh/.virtualenvs/angr/lib/python3.5/site-packages/claripy-8.18.10.25-py3.5.egg/claripy/ast/base.py", line 887, in ite_excavated
  File "/home/wgh/.virtualenvs/angr/lib/python3.5/site-packages/claripy-8.18.10.25-py3.5.egg/claripy/ast/base.py", line 831, in _excavate_ite
  File "/home/wgh/.virtualenvs/angr/lib/python3.5/site-packages/claripy-8.18.10.25-py3.5.egg/claripy/ast/base.py", line 831, in <listcomp>
  File "/home/wgh/.virtualenvs/angr/lib/python3.5/site-packages/claripy-8.18.10.25-py3.5.egg/claripy/ast/base.py", line 887, in ite_excavated
  File "/home/wgh/.virtualenvs/angr/lib/python3.5/site-packages/claripy-8.18.10.25-py3.5.egg/claripy/ast/base.py", line 831, in _excavate_ite
[...]

PR #113, opened 2018-11-05
Base: make _excavate_ite() non-recursive

danse-macabre:
Closes https://github.com/angr/claripy/issues/111.

I've also tried to change `_burrow_ite()` to be non-recursive, but failed to do so in a reasonable amount of time. Let me know if `_burrow_ite()` is also causing any troubles, and I'll revisit it later.


rhelmot:
Fantastic, thanks!

PR #117, opened 2018-11-25
General speed improvement in AST creation and simplification.

ltfish:
- Avoid re-scanning large args arrays again and again during AST
  creation.
- Make Bool.is_false() and Bool.is_true() faster.
- Use set instead of OrderedSet in several places for better
  performance.

### DIFF
--- a/claripy/backends/backend_concrete.py
+++ b/claripy/backends/backend_concrete.py
@@ -145,11 +145,22 @@ class BackendConcrete(Backend):
             raise UnsatError('concrete False constraint in extra_constraints')
         return self.convert(expr) == v
 
+    # Override Backend.is_true() for a better performance
+    def is_true(self, e, extra_constraints=(), solver=None, model_callback=None):
+        if e in {True, 1, 1.}:
+            return True
+        return type(e) is Base and e.op == "BoolV" and len(e.args) == 1 and e.args[0] is True
+    # Override Backend.is_false() for a better performance
+    def is_false(self, e, extra_constraints=(), solver=None, model_callback=None):
+        if e in {False, 0, 0.}:
+            return True
+        return type(e) is Base and e.op == "BoolV" and len(e.args) == 1 and e.args[0] is False
+
     #pylint:disable=singleton-comparison
     def _is_true(self, e, extra_constraints=(), solver=None, model_callback=None):
-        return e == True
+        raise NotImplementedError("This method should never be called since we override is_true().")
     def _is_false(self, e, extra_constraints=(), solver=None, model_callback=None):
-        return e == False
+        raise NotImplementedError("This method should never be called since we override is_false().")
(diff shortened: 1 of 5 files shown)


ltfish:
Evaluation:

```
def perf_boolean_and_simplification_0():
    # Create a gigantic And AST with many operands, one variable at a time
    bool_vars = [ claripy.BoolS("b%d" % i) for i in range(1500) ]
    v = bool_vars[0]
    for i in range(1, len(bool_vars)):
        v = claripy.And(v, bool_vars[i])


def perf_boolean_and_simplification_1():
    # Create a gigantic And AST with many operands, many variables at a time
    bool_vars = [ claripy.BoolS("b%d" % i) for i in range(500) ]
    v = bool_vars[0]
    for i in range(1, len(bool_vars)):
        if v.op == "And":
            v = claripy.And(*(v.args + (bool_vars[i] == False,)))
        else:
            v = claripy.And(v, bool_vars[i])


def perf():
    import timeit
    print(timeit.timeit("perf_boolean_and_simplification_0()",
                        number=10,
                        setup="from __main__ import perf_boolean_and_simplification_0"))
    print(timeit.timeit("perf_boolean_and_simplification_1()",
                        number=10,
                        setup="from __main__ import perf_boolean_and_simplification_1"))
```
On `master`:

```
$ python -O test_simplify.py
36.60717606698745
21.849817855021683
```

On `re/ast_speed_improvement`:

```
$ python -O test_simplify.py
19.754106904991204
3.9545626099861693
```

claripy/backends/backend_concrete.py
```diff
@@ -145,6 +145,21 @@ def _solution(self, expr, v, extra_constraints=(), solver=None, model_callback=N
...
 
+    # Override Backend.is_true() for a better performance
+    def is_true(self, e, extra_constraints=(), solver=None, model_callback=None):
+        if e in {True, 1, 1.}:
+            return True
+        if type(e) is Base and e.op == "BoolV" and len(e.args) == 1 and e.args[0] is True:
+            return True
+        return super().is_true(e, extra_constraints=extra_constraints, solver=solver,model_callback=model_callback)
+    # Override Backend.is_false() for a better performance
+    def ais_false(self, e, extra_constraints=(), solver=None, model_callback=None):
```
schieb:
```suggestion
    def is_false(self, e, extra_constraints=(), solver=None, model_callback=None):
```

Probably would have been caught, but just in case...

ltfish:
Note to (future) authors of AST canonicalization:  `OrderedSet` is a lot slower than `set()` in Python. In one of my test cases, we were spending 4.2 seconds in `OrderedSet.__add__()`, while we were only using about 0.2 seconds in `set()`. We should really look into fast OrderedSet alternatives if we want to use it in the future.

PR #118, opened 2018-12-08
Fix to account for cases where the first argument of a multi-argument expression is a constant

subwire:
Balancer was improperly giving up handling an expression, as it checked the second argument only for multivalued-ness and assumed the first argument was as well.

Fixes crazy jumptable bugs, among probably many other things.


ltfish:
Sorry, this is a duplicate of #114 . I cannot believe Yan hasn't merged my PRs for such a long time. Let me rebase them and then merge them in...

PR #119, opened 2018-12-18
Speed up the Z3 backend by using low-level Z3 APIs

ltfish:


claripy/backends/backend_z3.py
```diff
@@ -612,8 +612,9 @@ def _add(self, s, c, track=False):
         if track:
             for constraint in c:
                 name = str(hash(constraint))
-                self._hash_to_constraint[name] = constraint
-                s.assert_and_track(constraint, name)
+                if name not in self._hash_to_constraint.keys():
```
zardus: it'd be more performant to do `if name not in self._hash_to_constraint`, no?

claripy/backends/backend_z3.py
```diff
@@ -309,15 +309,15 @@ def FPS(self, ast):
     def FPV(self, ast):
         val = str(ast.args[0])
         sort = self._convert(ast.args[1])
-        if val in ("+oo", "+inf", "+Inf", 'inf'):
+        if val == "+oo" or val == "+inf" or val == "+Inf" or val == 'inf':
```
zardus: Why these changes, out of curiosity?

Phat3: Ops those are wrong changes which slipped in during the merge with the branch `wip/z3_updates`... Fixed now.

claripy/backends/backend_z3.py
Phat3: Yeap totally agree, my bad. Fixed now

ltfish:
I'm tempted to merge this branch, but there is still a few TODOs in the code. @zardus do you want to take a look at each of them and let me know what you think?

zardus:
It seems reasonable to merge this, and then keep working at further optimizations. Specifically, the TODOs that mention places that still require high-level APIs due to a lack of certainty of some variable types can be addressed by making BackendZ3's `convert` (or, more likely, AST operations themselves) stricter regarding those types (i.e., converting all ints to BVs and so forth). But that's a bit of a rabbit hole, so I think we should merge this and then keep moving forward.

ltfish:
Rebase does not work because of conflicts. I'm going to squash these commits. Great job, @Phat3 !

Issue #128, opened 2019-02-19
new String class shadows "replace" method


rhelmot:
String.replace should be renamed since ASTs have a .replace method which replaces elements in the tree.

Phat3:
Fixed in #129

PR #129, opened 2019-02-19
Removed replace method in ast.string class 

Phat3:
Resolved issue #128 and fixed a missing import in solver popen classes


rhelmot:
Are you sure... removing it is the best option? couldn't it just be renamed?

Phat3:
It is not used by anyone. I used to use it in the past in one of my experiment and I forgot to remove it during the merge. Now all the functions that operate on strings are directly called using class methods (e. g. `StrReplace`, `StrConcat`, etc.).

But yeah on second thought I agree that maybe it's better to leave it there renamed if someone wants to use it.

PR #131, opened 2019-02-20
make_like: Support simplifying expressions.

ltfish:
Now make_like() will try to simplify the given expression one more time.
By default it does not attempt to simplify expressions unless explicitly
specifying "simplify=True". This is to keep the compatibility with many
existing simplifiers that don't support multi-arg versions of certain
operations (e.g., Add and Or).

I'm also trying to smuggle in a new simplifier:
Simplifying `ZeroExt(A, ZeroExt(B, x))` to `ZeroExt(A + B, x)`.

### DIFF
--- a/claripy/ast/base.py
+++ b/claripy/ast/base.py
@@ -303,12 +303,27 @@ class Base:
     #            yield backend.convert(a)
 
     def make_like(self, *args, **kwargs):
+        op, expr_args = args
+        if kwargs.pop("simplify", False) is True:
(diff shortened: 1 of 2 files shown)


zardus:
@ltfish, were you happy with this solution in the end? I'm okay with it if it works for you.

ltfish:
Wait I did not realize that we never merge in this branch.

It worked for me and produced simpler ASTs.

PR #133, opened 2019-02-26
Optimizing the performance of concrete execution

ltfish:


ltfish:
The failing linting looks like a false positive in pylint. Ignored.

PR #138, opened 2019-05-15
Replace past.builtins.xrange for range

saullocarvalho:


claripy/ast/bool.py
```diff
@@ -182,7 +181,7 @@ def constraint_to_si(expr):
     satisfiable, replace_list = backends.vsa.constraint_to_si(expr)
 
     # Make sure the replace_list are all ast.bvs
-    for i in xrange(len(replace_list)):
+    for i in range(len(replace_list)):
```
ltfish: Please add `# pylint:disable=consider-using-enumerate` at the end of this line to stop pylint from complaining.

ltfish: See https://travis-ci.org/angr/claripy/jobs/533078142.

saullocarvalho: Is not better just use `enumerate` as suggested?

saullocarvalho: Ok!

ltfish: See line 188. We are modifying the list in place.

saullocarvalho: I missed that. I am adding the comment.
Thanks!

ltfish:
Thank you!

PR #142, opened 2019-05-29
Fixed two bugs with annotation arguments

orion--:
The first fixes the case where kwargs['annotations'] is None.

The second fixes the case with a default argument value, since 'None is False' is Falsey.


rhelmot:
Yep - these are good changes. Thanks!

PR #143, opened 2019-05-31
Let flattening simplifiers evaluate multiple concrete args together

rhelmot:
requires running simplifications during ite excavation to get tests to pass, but honestly that's not too bad of a deal tbh


ltfish:
Can you add test cases for this PR?

rhelmot:
I added some tests a while ago. Are you ok with these changes?

zardus:
They also look good to me, especially with the tests :-)

PR #145, opened 2019-06-07
Updated backend Z3 to work with the newest Z3 version

Phat3:
Signed-off-by: Sebastiano Mariani <mariani.sebastiano@gmail.com>

### DIFF
--- a/claripy/backends/backend_z3.py
+++ b/claripy/backends/backend_z3.py
@@ -51,6 +51,8 @@ solve_count = 0
 
 supports_fp = hasattr(z3, 'fpEQ')
 
+# you can toggle this flag if you want. I don't think it matters
+#z3.set_param('rewriter.hi_fp_unspecified', 'true')
 
 #
 # Utility functions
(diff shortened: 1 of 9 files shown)


claripy/backends/backend_z3.py
```diff
@@ -1,6 +1,7 @@
 
 import os
 import z3
+import gc
```
ltfish: Why do we need `gc`?

claripy/backends/backend_z3.py
```diff
@@ -46,6 +47,14 @@ def _add_memory_pressure(p):
 
 supports_fp = hasattr(z3, 'fpEQ')
 
+# Hack to get a reference to z3libs. Thanks Rhelmot!
```
ltfish: `Rhelmot` -> `rhelmot`

@rhelmot Correct me if I'm wrong.

ltfish:
Thanks for doing this! But I feel you are committing some local workarounds, which should not be part of this patch. Can you please check this PR again?

claripy/backends/backend_z3.py
rhelmot: @Phat3 you can get rid of the gc hack and replace the usage of the libz3 object with this: https://github.com/Z3Prover/z3/pull/2143

PR #146, opened 2019-06-10
Fix docstrings

saullocarvalho:


claripy/strings.py
```diff
@@ -100,17 +108,21 @@ def StrSuffixOf(suffix, input_string):
...
     return re.match(r'.*' + suffix.value + '$', input_string.value) is not None
 
+
 def StrIndexOf(input_string, substring, startIndex, bitlength):
     """
-    Return True if the concrete value of the input_string ends with suffix
-    otherwise false.
+    Return a concrete BitVector of bitlength size. Its value is the index of
+    the first occurrence of substring searching from startIndex if found
+    otherwise is -1.
```
rhelmot: This is a very backwards way of saying this. The type of the result isn't as important as its meaning. Also, the english is broken. Try: `Find the first occurance at or after the start index of the given substring in the input string, or -1 if it is not found.`. The return type is specified in the `:return:` field.

claripy/strings.py
```diff
@@ -100,17 +108,21 @@ def StrSuffixOf(suffix, input_string):
...
     :param input_string: the string we want to check
     :param substring: the substring we want to find the index
     :param startIndex: the index to start searching at
-    :param bitlength: bitlength of the bitvector representing the index of the substring
+    :param bitlength: bitlength of the bitvector representing the index of the
+                      substring
 
-    :return BVV: index of the substring in bit-vector representation or -1 in bitvector representation
+    :return BVV: index of the substring in bitvector representation or -1 in
+                 bitvector representation
```
rhelmot: BVV is not a type; you want to say BV here. Also "bitvector representation" is a bit purple-prose for API documentation. Just say it's a bitvector.

claripy/strings.py
```diff
@@ -39,12 +42,13 @@ def StrExtract(high, low, str_val):
...
+    (replace ONLY the first occurrence of the pattern)
 
     :param initial_string: string in which the pattern needs to be replaced
-    :param pattern_to_be_replaced: substring that has to be replaced inside initial_string
-    :param replacement_pattern: pattern that has to be inserted in initial_string to replace
-                                 pattern_to_be_replaced
+    :param pattern_to_be_replaced: substring that has to be replaced inside
+                                   initial_string
+    :param replacement_pattern: pattern that has to be inserted in
+                                initial_string to replace pattern_to_be_replaced
```
rhelmot: The way we usually do indentation in docstrings is that all the text after the names gets indented to the same level across all the lines. If that's too obnoxious you can stick a newline after the `:param name:`.

Also, we prefer to wrap our text at 100 cols in this project.

claripy/strings.py
rhelmot: I realize you're just copying the style of what was already there, oops. You should change it, the way it was is bad!

claripy/strings.py
```diff
@@ -120,37 +132,44 @@ def StrIndexOf(input_string, substring, startIndex, bitlength):
...
         return BVV(int(input_string.value), bitlength)
     except ValueError:
         return BVV(-1, bitlength)
 
+
 def StrIsDigit(input_string):
     """
-    Returns True if the concrete value of the input_string is a digit otherwise returns False
+    Returns True if the concrete value of the input_string is a digit otherwise
+    returns False
```
rhelmot: This is basically a verbose copy of the `:return:` field. Plus, it's wrong. Try "Determine whether the given string is entirely numeric"

saullocarvalho:
I've made all the requested changes and I've turned the docstrings less verbose. I hope it is better!

claripy/strings.py
```diff
@@ -2,31 +2,36 @@
...
 
     def __repr__(self):
         return 'StringV(%s)' % (self.value)
 
+
 def StrConcat(*args):
     """
-    Create a concrete version of the concatenated string
-    :param args: List of string that has to be concatenated
+    Create a concrete version of the concatenated string.
```
rhelmot: ...agh. I wish I had reviewed the PR that added string more carefully. All of these functions don't create concrete anything, they are merely the operations implemented for the concrete String objects.

saullocarvalho:
I hope that adding the parameters to the docstrings make them more understandable.

rhelmot:
Thanks!!

Issue #19, opened 2016-06-28
Restore docstrings lost during the solver redesign


zardus:
We lost all the docstrings on `claripy.Frontend`. It would be great to restore them, and propagate them through the inheritance tree to the various Solvers (in `claripy/solvers.py`).

saullocarvalho:
I am interested in helping. What can I do?

zardus:
Thanks to @saullocarvalho's heroic efforts, this three-year-old issue can be finally closed!

:fireworks:

PR #149, opened 2019-06-12
Add docstrings to claripy.Frontend

saullocarvalho:
As I could not find any more lost docstrings, I've written these ones based on experimentation and code reading. I hope that I've got them right.

More fixes to issue #19


zardus:
Brilliant. I'm going to merge this and then add a bit more discussion on `is_true` and `is_false`, since those often cause confusion. Thanks!

Issue #141, opened 2019-05-28
Need a concept of operator precedence in claripy op rendering


rhelmot:
It would be nice to extend the way `Base._op_repr` worked to make it so that it put parentheses only where they needed to go.

saullocarvalho:
I am interested in helping. What can I do?

rhelmot:
So we need:

- add some sort of mapping in claripy from op string which has a symbol associated (e.g. `__add__` has the + symbol associated) to precedence level, and some understanding of the associativity of each level, a la https://en.cppreference.com/w/c/language/operator_precedence
- change the ast repr process so that it only adds parentheses around an operation if its parent operation is of a higher (smaller number in the linked chart) precedence or if it violates the associativity rules
- tests

The ast repr process is currently a dizzyingly complicated iterated algorithm instead of a recursive algorithm because at one point someone contributed a change switching all our recursive AST processing code into iterated stack-based algorithms. I would be okay if you rewrote the AST rendering algorithm from the ground up as a recursive algorithm again; I could do the conversion to the iterated form myself.

PR #151, opened 2019-06-13
Fix issue #141 

saullocarvalho:
Fix issue #141 .

I've converted the iterated AST processing algorithm to a recursive one, and I've modified `shallow_repr` to consider infix operation precedence level.

### DIFF
--- a/claripy/ast/base.py
+++ b/claripy/ast/base.py
@@ -430,7 +430,7 @@ class Base:
         else:
             return self.shallow_repr(max_depth=max_depth, explicit_length=explicit_length, inner=inner)
 
-    def shallow_repr(self, max_depth=8, explicit_length=False, details=LITE_REPR, inner=False, parent_prec=15):
+    def shallow_repr(self, max_depth=8, explicit_length=False, details=LITE_REPR, inner=False, parent_prec=15, left=True):
         """
         Returns a string representation of this AST, but with a maximum depth to
         prevent floods of text being printed.
@@ -460,11 +460,11 @@ class Base:
         # if operation is not in op_precedence, assign the "least operation precedence"
         op_prec = operations.op_precedence[op] if op in operations.op_precedence else 15
 
-        args = [arg.shallow_repr(next_max_depth, explicit_length, details, True, op_prec) \
-                if isinstance(arg, Base) else arg for arg in args]
+        args = [arg.shallow_repr(next_max_depth, explicit_length, details, True, op_prec, idx == 0) \
+                if isinstance(arg, Base) else arg for idx, arg in enumerate(args)]
 
-        parent_higher_prec = parent_prec < op_prec
-        inner_repr = self._op_repr(op, args, inner, length, details, parent_higher_prec)
+        prec_diff = parent_prec - op_prec
+        inner_repr = self._op_repr(op, args, inner, length, details, prec_diff, left)
 
         if not inner:
             return "<{} {}>".format(self._type_name(), inner_repr)
@@ -472,7 +472,7 @@ class Base:
             return inner_repr
 
     @staticmethod
-    def _op_repr(op, args, inner, length, details, parent_higher_prec):
+    def _op_repr(op, args, inner, length, details, prec_diff, left):
         if details < Base.FULL_REPR:
             if op == 'BVS':
                 extras = []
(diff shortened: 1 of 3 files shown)


zardus:
@rhelmot, do you want to take a look to see if this is what you envisioned? I'm ready to merge this otherwise as a solid start to this concept!

zardus:
(whenever you have a moment, ofc)

tests/test_ast.py
```diff
@@ -0,0 +1,24 @@
+import nose.tools
+
+import claripy
+
+
+def test_lite_repr():
+    one = claripy.BVV(1, 8)
+    two = claripy.BVV(2, 8)
+    a = claripy.BVS('a', 8)
+    b = claripy.BVS('b', 8)
```
rhelmot: you can avoid the regex assert (please avoid the regex assert) by either templating the target string or using `explicit_name=True` in the BVS constructor.

claripy/ast/base.py
```diff
@@ -197,10 +198,11 @@ def _calc_hash(op, args, keywords):
...
         Calculates the hash of an AST, given the operation, args, and kwargs.
 
-        :param op:          The operation.
-        :param args:        The arguments to the operation.
-        :param keywords:    A dict including the 'symbolic', 'variables', and 'length' items.
-        :returns:           a hash.
+        :param op:                  The operation.
+        :param args:                The arguments to the operation.
+        :param keywords:            A dict including the 'symbolic', 'variables', and 'length' items.
+
```
rhelmot: I would prefer not to have a blank line in between the parameters and the return value. rst will treat the list as two separate lists when the newline is inserted.

tests/test_ast.py
saullocarvalho: I didn't know about `explicit_name=True` use. Thanks for the information!
I've just made the change.

claripy/ast/base.py
saullocarvalho: Blank lines removed!
I am going to fix it on angr-docs. The blank line between parameters and the return value seems expected.

saullocarvalho:
I've added tests to check left-to-right associativity.

tests/test_ast.py
```diff
@@ -6,19 +6,31 @@
...
 
 
+def test_associativity():
+    x = claripy.BVS('x', 8, explicit_name=True)
+    y = claripy.BVS('y', 8, explicit_name=True)
+    z = claripy.BVS('z', 8, explicit_name=True)
+    w = claripy.BVS('w', 8, explicit_name=True)
+
+    nose.tools.assert_equal((x - y - z - w).shallow_repr(),
+                            (x - (y - (z - w))).shallow_repr())
```
rhelmot: this test is not correct! `x - y - z - w` is the same as `+x + -y + -z + -w`, while `x - (y - (z - w))` is the same as `+x + -y + +z + -w`

tests/test_ast.py
```diff
@@ -6,19 +6,31 @@
...
-    nose.tools.assert_regex((a + one * b + two).shallow_repr(),
-                            r'\<BV8 __add__\(a_\d+_8, 1 \* b_\d+_8, 2\)\>')
-    nose.tools.assert_regex(((a + one) * (b + two)).shallow_repr(),
-                            r'\<BV8 \(a_\d+_8 \+ 1\) \* \(b_\d+_8 \+ 2\)\>')
-    nose.tools.assert_regex((a * one + b * two).shallow_repr(),
-                            r'\<BV8 a_\d+_8 \* 1 \+ b_\d+_8 \* 2\>')
-    nose.tools.assert_regex(((one + a)*(two + b)+(two + a)*(one + b)).shallow_repr(),
-                            r'\<BV8 \(1 \+ a_\d+_8\) \* \(2 \+ b_\d+_8\) \+ \(2 \+ a_\d+_8\) \* \(1 \+ b_\d+_8\)\>')
+    nose.tools.assert_equal((a + one * b + two).shallow_repr(),
+                            '<BV8 __add__(a, 1 * b, 2)>')
```
rhelmot: can we make it so that this renders as `a + 1 * b + 2`?

tests/test_ast.py
saullocarvalho: Oops! You're right. I've missed that detail. I am gonna fix it.

tests/test_ast.py
saullocarvalho: Yes. We can.

saullocarvalho: The infix operator with more than 2 args is full represented. Even when using `details == Base.LITE_REPR`. I've changed this.

tests/test_ast.py
saullocarvalho: I've misunderstood your previous comment. Thanks for pointing the mistake. I've fixed the associativity tests and the code to handle it correctly.

rhelmot:
absolutely beautiful. thank you!

PR #152, opened 2019-06-17
Add a debug mode to allow users to turn off certain checks for better performance.

ltfish:


zardus:
The way you're doing `from debug import _DEBUG` means that only the value of `_DEBUG` at import time will matter, and `set_debug` won't actually do anything, will it?

ltfish:
Fixed!

zardus:
Looks good to me!

PR #154, opened 2019-06-17
Add more fast paths.

ltfish:
- Don't cache BVVs. Just create them instead. It's faster.
- BVVs don't support annotations any more.
- Run eager evaluation on concrete operations sooner.
- If no annotations exist, return sooner.

### DIFF
--- a/claripy/ast/base.py
+++ b/claripy/ast/base.py
@@ -88,7 +88,7 @@ class Base:
     MID_REPR=1
     FULL_REPR=2
 
-    def __new__(cls, op, args, add_variables=None, hash=None, **kwargs): #pylint:disable=redefined-builtin
+    def __new__(cls, op, args, add_variables=None, hash_=None, **kwargs): #pylint:disable=redefined-builtin
         """
         This is called when you create a new Base object, whether directly or through an operation.
         It finalizes the arguments (see the _finalize function, above) and then computes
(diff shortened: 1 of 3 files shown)


claripy/ast/base.py
```diff
@@ -112,8 +112,36 @@ def __new__(cls, op, args, add_variables=None, hash=None, **kwargs): #pylint:dis
 
         a_args = args if type(args) is tuple else tuple(args)
 
-        # initialize the following properties: symbolic, variables and errored
+        # fast path: do not hash or cache BVV
```
zardus: There are a lot of places in angr where we use `is` to reason about equality of BVVs, especially (but not only) in the simplifiers. This will break that. If we want this route, we should implement something like `ast.is_same` to replace using `is`, though that might slow things down significantly...

rhelmot: In the next angr major version we should have == check structural equality (with fast paths of course) and add a .equals for the comparison operation.

zardus: That makes me nervous... Having an `==` that can't return a BoolS will inject the sort of confusion that happens over `is_true` and `is_false`, but to everything, and needing to call `.equals` a) gives me Java PTSD and b) means `a.equals(b)` could AttributeError out when `a` is an int, whereas `a == b` currently won't.

PR #157, opened 2019-08-02
Cache abstraction info for constraints when tracking is enabled

rhelmot:
Otherwise, we will try to abstract z3's normalized version of the constraints. This is undesirable for my usecase.


rhelmot:
(this is not related to the garbage collection stuff I was complaining about earlier)

PR #159, opened 2019-10-24
WIP: Type Annotations for claripy 

fmagin:
Discussion in angr/pyvex#198

No actual annotations in here yet, but I will probably need some in the near future

### DIFF
--- a/claripy/py.typed
+++ b/claripy/py.typed
@@ -1,0 +1,1 @@
+partial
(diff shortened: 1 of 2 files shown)


fmagin:
Functionally superseded by https://github.com/angr/claripy/pull/255 anyway, which actually adds some annotations and not just the `py.typed` file

PR #161, opened 2019-12-19
Implement operations from ast/strings.py in the z3 backend

embg:
This patch implements all the claripy string operations (except IsDigit) in the z3 backend. It also adds a new string operation, UnitStr, which creates a single-char string from a bitvector. I am using this code for experiments with making the simprocedures for `<string.h>` functions faster by passing constraints to z3's powerful string solver, `z3str3`. Any feedback on the changes would be welcome. This is my first PR to claripy, so apologies in advance for any errors or misunderstandings :)

Edit: I have tested `StrIndexOf` with `StringS` and `StringV` values on my machine and it works fine (that's the main operation I need for my work in angr). The other operations have not been tested, but should work based on how similar their implementations are to `StrIndexOf`. If anyone is actually interested in using the other functions, I'd be happy to write unit tests for all of them.

Edit: It should also be noted that my code uses z3's definition of a string as a sequence of 8-bit BVs. If BVs of different width are used much of the code will work fine, but some weird errors may crop up. In particular this will limit the usage of the code for wide-char variants of `<string.h>` functions.

### DIFF
--- a/claripy/ast/strings.py
+++ b/claripy/ast/strings.py
@@ -154,6 +154,7 @@ StrIndexOf = operations.op("StrIndexOf", (String, String, BV, int), BV, calc_len
 StrToInt = operations.op("StrToInt", (String, int), BV, calc_length=operations.strtoint_bv_size_calc, bound=False)
 IntToStr = operations.op("IntToStr", BV, String, calc_length=operations.int_to_str_length_calc, bound=False)
 StrIsDigit = operations.op("StrIsDigit", String, Bool, bound=False)
+UnitStr = operations.op("UnitStr", BV, String, bound=False) # convert BV to single-char string
 
 # Equality / inequality check
 String.__eq__ = operations.op('__eq__', (String, String), Bool)
@@ -173,3 +174,4 @@ String.StrIndexOf = staticmethod(StrIndexOf)
 String.StrToInt = staticmethod(StrToInt)
 String.StrIsDigit = staticmethod(StrIsDigit)
 String.IntToStr = staticmethod(IntToStr)
+String.UnitStr = staticmethod(UnitStr)
--- a/claripy/ast/strings.py
+++ b/claripy/ast/strings.py
@@ -152,9 +152,9 @@ StrPrefixOf = operations.op("StrPrefixOf", (String, String), Bool, bound=False)
 StrSuffixOf = operations.op("StrSuffixOf", (String, String), Bool, bound=False)
 StrIndexOf = operations.op("StrIndexOf", (String, String, BV, int), BV, calc_length=operations.strindexof_bv_size_calc, bound=False)
 StrToInt = operations.op("StrToInt", (String, int), BV, calc_length=operations.strtoint_bv_size_calc, bound=False)
-IntToStr = operations.op("IntToStr", BV, String, calc_length=operations.int_to_str_length_calc, bound=False)
-StrIsDigit = operations.op("StrIsDigit", String, Bool, bound=False)
-UnitStr = operations.op("UnitStr", BV, String, bound=False) # convert BV to single-char string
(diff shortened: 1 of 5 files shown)


claripy/ast/strings.py
```diff
@@ -151,6 +154,7 @@ def StringV(value, length=None, **kwargs):
 StrToInt = operations.op("StrToInt", (String, int), BV, calc_length=operations.strtoint_bv_size_calc, bound=False)
 IntToStr = operations.op("IntToStr", BV, String, calc_length=operations.int_to_str_length_calc, bound=False)
 StrIsDigit = operations.op("StrIsDigit", String, Bool, bound=False)
+UnitStr = operations.op("UnitStr", BV, String, bound=False) # convert BV to single-char string
```
rhelmot: The meaning of having a single type as opposed to a tuple of types in the input argument is that it accepts a variadic number of arguments of that type. Is that what you want here? The other operations may be specified wrong... I don't know their semantics off the top of my head.

claripy/backends/backend_z3.py
```diff
@@ -156,6 +156,20 @@ def __init__(self, reuse_z3_solver=None, ast_cache_size=10000):
...
+        self._op_raw['StrLen'] = self._op_raw_StrLen
+        self._op_raw['StrReplace'] = self._op_raw_StrReplace
+        self._op_raw['StrContains'] = self._op_raw_StrContains
+        self._op_raw['StrPrefixOf'] = self._op_raw_StrPrefixOf
+        self._op_raw['StrSuffixOf'] = self._op_raw_StrSuffixOf
+        self._op_raw['StrIndexOf'] = self._op_raw_StrIndexOf
+        self._op_raw['StrToInt'] = self._op_raw_StrToInt
+        self._op_raw['IntToStr'] = self._op_raw_IntToStr     
+        self._op_raw['UnitStr'] = self._op_raw_UnitStr 
+
```
rhelmot: Is this necessary? It looks like there's code up above to loop over the class's methods with this naming scheme and automatically set them into this dictionary.

claripy/backends/backend_z3.py
```diff
@@ -1265,7 +1307,7 @@ def _op_raw_SGT(a, b):
     @condom
     def _op_raw_SGE(a, b):
         return a >= b
-
+
```
rhelmot: remove trailing whitespace

embg:
> This is cool stuff - but can I ask how exactly you're actually integrating string types into the byte-addressed memory model?

Thanks! The idea is to build a string where each character is constrained equal to a byte in memory. This can be accomplished by starting with an empty string and `Concat`ing a series of `Unit` strings (single-char strings) for each byte. Then we can get, for example, `strlen` as an integer by doing `IndexOf(s, '\x00')` and converting the integer to a bitvector using `Int2BV`.

claripy/ast/strings.py
embg: OK, I will need to fix a few of the implementations (which are not variadic and should be) and fix my signature for `UnitStr`. I'll make a commit with those changes tomorrow.

claripy/backends/backend_z3.py
embg: Correct, I'll delete those in my commit tomorrow

rhelmot:
Is this still WIP? I can merge it if you like. The CI failures are spurious.

embg:
> Is this still WIP? I can merge it if you like. The CI failures are spurious.

I just marked it WIP because I haven't finished the angr code which calls this code and wanted to avoid opening a second PR with bugfixes from that work. If you're ok with a possible second PR then it should be ready to merge :) thanks for reviewing

rhelmot:
this all seems self-explanitory to the point that I'm not super worried about issues popping up. If you do find some we always appreciate fixes :)

PR #164, opened 2020-02-25
Make composite solver more efficient in the presence of many independent constraint sets

zardus:
Until now, the composite solver would loop through all of its children to check if it is satisfiable. Even if each check hit the child's cache, looping through the children would make this process slow. Now, the composite solver will only check solvers that it does not already know to be satisfiable.

One specific side-effect of this is that if someone accesses the children directly, and adds constraints to them, the composite solver won't know to re-check. There were already other parts of the composite solver that would break in that scenario, though, so I think that's okay.


ltfish:
Ref: https://github.com/angr/angr/issues/1612

PR #167, opened 2020-03-05
Save and restore _finalize attribute when pickling.

edmcman:


rhelmot:
This looks like a good change, and the test failures are also present on master. Thanks!

PR #170, opened 2020-04-16
fix missing extra_constraints in CompositeFrontend.unsat_core()

gkso:
In the CompositeFrontend.unsat_core function, extra_constraints was not passed to child solvers. Therefore it cannot get an unsat core if it's the extra_constraints causing the unsatisfiability.


rhelmot:
good catch. Thanks!

PR #172, opened 2020-04-23
Implement more boolean operations in BackendConcrete

rhelmot:
Somehow in the linked issue, BackendConcrete was getting hung up on stuff like Xor(True, Xor(True, True)). This resolves that.


rhelmot:
I've seen the failing test on master, though I can't reproduce it locally. I believe it has to do with gc patterns...

PR #177, opened 2020-05-08
Annotation fixes

AOS002:
- If no simplifications are done when eliminatable=Fale & relocatable=True, then the annotations still need to pass from the args to the new AST
- The annotations were not being passed through to the results for the claripy.If()

### DIFF
--- a/claripy/ast/base.py
+++ b/claripy/ast/base.py
@@ -273,7 +273,7 @@ class Base:
             ))).keys()
 
             self.annotations = tuple(itertools.chain(
-                itertools.chain.from_iterable(a.annotations for a in ast_args),
+                itertools.chain.from_iterable(a._relocatable_annotations for a in ast_args),
                 tuple(a for a in self.annotations)
             ))
 
--- a/claripy/ast/base.py
+++ b/claripy/ast/base.py
@@ -272,11 +272,6 @@ class Base:
                 tuple(a for a in self.annotations if not a.eliminatable and a.relocatable)
             ))).keys()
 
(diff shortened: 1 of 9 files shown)


claripy/ast/base.py
```diff
@@ -272,6 +272,11 @@ def __a_init__(self, op, args, variables=None, symbolic=None, length=None, simpl
                 tuple(a for a in self.annotations if not a.eliminatable and a.relocatable)
             ))).keys()
 
+            self.annotations = tuple(itertools.chain(
```
zardus, line 275: I might be missing something, but doesn't the above already catch eliminatable=False, relocatable=True annotations?

AOS002: Yes that is correct, but they do not pass on the annotations from the `ast_args` to the resultant AST they only set the `self._uneliminatable` & `self._relocatable`. I am assuming that if `relocatable=True` and `eliminatable=False/True`, then the annotations from the arguments should be relocated to the annotations of the new AST(i.e. `self.annotations`).

zardus:
I forgot about this until now! Fish convinced me that it is correct. Are we good to merge?

AOS002:
I think we are good to merge! although I see that there is one test that's failing. The same test is failing on the most recent commit as well, so I don't think it should be because of the changes I made.

rhelmot:
I'm hitting merge on this. The CI run has vanished into the ether, but based on what was said I'm gonna assume it was test_unsat_core failing, which is known nondeterministic.

PR #181, opened 2020-07-20
Change ite_dict to use binary search tree

Timeroot:
Currently, `ite_dict` turns a dictionary lookup into a claripy expression by using a linear series of if-then-else statements. This creates expression trees whose depths are equal to the size of the dict. For a dict with, say, 200 elements, this creates a RecursionError in claripy. This came up recently for me when [hooking an instruction that was most easily implemented by a table](https://ohaithe.re/post/624142953693249536/uiuctf-2020-cricket32) of 256 elements, and claripy crashed.

This PR replaces `ite_dict` to use a binary search tree when the dict gets biggish. It would be risky to do this for `ite_cases`, since there they might be an assumption that it tries the cases in the given order, in the event that two cases are both true. But for a dict, only one case can be true, so this binary searching won't change the functionality.

### DIFF
--- a/claripy/ast/bool.py
+++ b/claripy/ast/bool.py
@@ -161,16 +161,29 @@ def is_false(e, exact=None): #pylint:disable=unused-argument
     l.debug("Unable to tell the truth-value of this expression")
     return False
 
-def ite_dict(i, d, default):
-    """
-    Return an expression of if-then-else trees which expresses a switch tree
-
-    :param i: The variable which may take on multiple values affecting the final result
-    :param d: A dict mapping possible values for i to values which the result could be
-    :param default: A default value that the expression should take on if `i` matches none of the keys of `d`
-    :return: An expression encoding the result of the above
-    """
-    return ite_cases([ ((i.ast if type(i) is ASTCacheKey else i) == c, v) for c,v in d.items() ], default)
+#For large tables, ite_dict that uses a binary search tree instead of a "linear" search tree.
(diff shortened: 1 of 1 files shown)


rhelmot:
oh hai

this is good code! I was going to complain that it assumes that the keys of the dict are ints, but it looks like the original code makes that assumption anyawy. In that case, the only thing you need to fix is:

```
LINT FAILURE: claripy/ast/bool.py regressed to 9.65/10.00
... claripy/ast/bool.py:168:0: C0303: Trailing whitespace (trailing-whitespace)
... claripy/ast/bool.py:177:0: C0303: Trailing whitespace (trailing-whitespace)
... claripy/ast/bool.py:181:0: C0303: Trailing whitespace (trailing-whitespace)
... claripy/ast/bool.py:187:0: C0303: Trailing whitespace (trailing-whitespace)
... claripy/ast/bool.py:191:0: C0303: Trailing whitespace (trailing-whitespace)
```

ltfish:
Eventually, after so many years... @zardus 

I feel I have implemented this before though. Probably in some exotic local claripy branch I have :/

Thanks a lot!

Timeroot:
Fixed the lints. Just curious, is there anything it would make sense to pass into the dict that *isn't* ints? (...strings...?)

rhelmot:
The point of an ASTCacheKey is that it represents an AST and can be used as a dict key (i.e. the equality function works non-pathologically), so theoretically you could pass one of those in. But I guess that's what `ite_cases` is for.

PR #192, opened 2020-12-20
Moved version to /VERSION and updated setup.py as needed

zwimer:


rhelmot:
this will mess up our release scripts. I'm assuming you need this for the C++ build? Can you have the version number passed into the C++ build through an environment variable that setup.py can set?

twizmwazin:
@zwimer and I discussed this, the changes to the release scripts would not be major. It would just involve adding claripy to [this(https://github.com/angr/ci-settings/blob/master/scripts/make_release_commits.sh#L21) clause and moving it to the top of the if-else chain.

twizmwazin (approved):

PR #196, opened 2021-01-05
Removed dead code from operations.py

zwimer:
This removes the 'Reversed' operation and the set it was defined in from operations.py, as nothing referenced either in all of claripy.


ltfish:
Also, are `length_none_operations`, `length_change_operations`, and `length_new_operations` used anywhere? If not, let's get rid of them, too.

ltfish (approved):

ltfish:
Merged. Thanks!

PR #206, opened 2021-02-24
Fix/remove nonexistent ops

zwimer:
These five ops are un-implemented throughout claripy, _at least as ops_. This PR is related to the conversation @ltfish, @zardus, and I had about these ops and how they should be removed.

### DIFF
--- a/claripy/backends/backend_vsa.py
+++ b/claripy/backends/backend_vsa.py
@@ -58,9 +58,6 @@ class BackendVSA(Backend):
         self._make_expr_ops(set(expression_set_operations), op_class=self)
         self._make_raw_ops(set(backend_operations_vsa_compliant), op_module=BackendVSA)
 
-        self._op_raw['StridedInterval'] = BackendVSA.CreateStridedInterval
-        self._op_raw['ValueSet'] = ValueSet.__init__
-        self._op_raw['AbstractLocation'] = AbstractLocation.__init__
         self._op_raw['Reverse'] = BackendVSA.Reverse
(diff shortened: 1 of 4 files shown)


zardus:
Looks like "identical" was used in some context...

zwimer:
Where?

It isn't defined anywhere in the claripy repo at least. If you `grep -r Identical` the repo you will find:
`claripy/backends/backend_z3.py:        self._op_raw['Identical'] = self._identical`
`claripy/operations.py:not_invertible = {'Identical', 'union'}`
Where the `_identical` function is 
```
def _identical(self, a, b):
         return a.eq(b)
```

The op isn't defined for any type though, so it could only be constructed by someone making an AST node with the op and arguments manually. (i.e. there is no `BV.Identical = operations.op(` line or anything like that).

Is the op AST node manually created someone else / in some other repo?

zardus:
Can you look at the failing testcases and figure out where that's coming
from?

zwimer:
It seems the identical function was tested, but the Identical op does not seem to have been used. I put the function back, but left the Op removed. If this still fails I will investigate why.

zwimer:
@zardus That fixed it. The 'new' errors are linting errors that are distinct from code changed in this PR.

Now that these ops are no longer being supported, we may want to look into the tooling around them. For example, the linter is complaining about `AbstractLocation` being imported but not used. This is possibly intentional, just something I figured was worth mentioning.

ltfish (approved):

PR #211, opened 2021-03-17
Fix lint errors

zwimer:
Fixes the lint errors preventing 205 from being merged.

### DIFF
--- a/claripy/ast/bv.py
+++ b/claripy/ast/bv.py
@@ -1,4 +1,3 @@
-import binascii
 import logging
 import numbers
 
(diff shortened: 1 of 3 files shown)


zwimer:
@rhelmot The only lint errors (that I saw) that I did not fix were these:
```
... claripy/ast/bv.py:89:11: E1101: Instance of 'NotImplemented' has no 'size' member (no-member)
... claripy/ast/bv.py:90:16: E1101: Instance of 'NotImplemented' has no 'zero_extend' member (no-member)
... claripy/ast/bv.py:90:34: E1101: Instance of 'NotImplemented' has no 'size' member (no-member)
... claripy/ast/bv.py:109:11: E1101: Instance of 'NotImplemented' has no 'size' member (no-member)
... claripy/ast/bv.py:110:16: E1101: Instance of 'NotImplemented' has no 'zero_extend' member (no-member)
... claripy/ast/bv.py:110:34: E1101: Instance of 'NotImplemented' has no 'size' member (no-member)
```
Though the line numbers should all be decremented by one. These seem like serious errors; though my assumption is that the linter is unaware of something that make these false positives. I didn't see any bug report open about these so I will try fixing them in a second commit later and see if CI passes or not.

rhelmot:
Yes, that is a known issue with pylint. I don't believe anyone has ever seriously attempted to track down its cause, though.

PR #218, opened 2021-05-11
Pin pysmt to remove old vendored six

twizmwazin:


rhelmot:
This is so weird. I had this pinned at one point but then I unpinned it because I thought they fixed that issue? Weird.

PR #219, opened 2021-05-20
Implement signed min and max

rhelmot:
:)

### DIFF
--- a/claripy/backends/backend_z3.py
+++ b/claripy/backends/backend_z3.py
@@ -830,11 +830,16 @@ class BackendZ3(Backend):
         return result_values
 
     @condom
-    def _min(self, expr, extra_constraints=(), solver=None, model_callback=None):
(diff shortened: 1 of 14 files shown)


claripy/backends/backend_z3.py
```diff
@@ -917,7 +930,7 @@ def _max(self, expr, extra_constraints=(), solver=None, model_callback=None):
             # TODO: is this assumption correct?
             # here it's not safe to call directly the z3 low level API since it might happen that the argument is an
             # integer and not a BV
-            solver.add(z3.UGT(expr, middle), z3.ULE(expr, hi))
```
rhelmot, line 920: One strange thing I noticed while working on this: it's weird how max uses GT and LE while min uses GE and LE. Is there a reason for this?

zardus:
There have been so many off-by-one errors in that code that it’s possible
that it was on purpose. I think a binary search is one of the hardest
algorithms in CS to actually implement without bugs :-)

PR #222, opened 2021-06-01
rm StrExtract

zwimer:
StrExtract is not used much as far as I can tell; is buggy / incorrect, and depends on z3 code that is buggy / has incorrect documentation. Furthermore, it is just an alias of SubStr

I replaced usages of StrExtract with their equivalent SubStr calls; so this should not affect runs much; however, since StrExtract was buggy, the call to SubStr (since I converted the arguments) likely is too.

### DIFF
--- a/claripy/ast/strings.py
+++ b/claripy/ast/strings.py
@@ -42,7 +42,7 @@ class String(Bits):
             # Because we are indexing from the end, what was high becomes low and vice-versa
             high_str_idx = self.string_length - 1 - low
             low_str_idx = self.string_length - 1 - high
-            return StrExtract(low_str_idx, high_str_idx + 1 - low_str_idx, self)
+            return StrSubstr(high_str_idx + 1 - low_str_idx, 2*low_str_idx - high_str_idx, self)
         else:
             raise ValueError("Only slices allowed for string extraction")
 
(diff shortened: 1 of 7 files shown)


claripy/ast/strings.py
```diff
@@ -42,7 +42,7 @@ def __getitem__(self, rng):
             # Because we are indexing from the end, what was high becomes low and vice-versa
             high_str_idx = self.string_length - 1 - low
             low_str_idx = self.string_length - 1 - high
-            return StrExtract(low_str_idx, high_str_idx + 1 - low_str_idx, self)
+            return StrSubstr(high_str_idx + 1 - low_str_idx, 2*low_str_idx - high_str_idx, self)
```
zwimer, line 45: This is almost certainly a bug; but it should be the same bug that `StrExtract` was hiding. We should fix it though, either now or in another PR.

rhelmot: Fix it now.

zwimer: Fixed, though please double check. I tested it as such:
```
import claripy
from z3 import *
def f(x,y):
  lambda x,y: print(z3.simplify(claripy.backends.z3.convert(claripy.StringV("abcdef")[x:y])))
f(31:16) # Should be (in bytes) [4:2]; which is start index: 1, count = 3, which is: 'bcd'.
```

rhelmot (approved):

PR #231, opened 2021-07-13
Fallback to hashlib if _md5 is unavailable

recursiveMake:
Some Python distributions do not include the _md5 module.


ltfish:
Thanks!

PR #235, opened 2021-09-02
Min/max do not bother solving model

zwimer:
Min/max should just use the end result of the binary search rather than solving a model every step, adding the result to a set, then taking the extrema of the set which should just be the end of the binary search in the first place.


rhelmot:
hm. well that's a weird way to write this code that I'm not sure I've ever actually questioned.

I can't see any way this could possibly break something unless there's a discrepancy between z3 and python's numeric comparisons (signedness?), and even then I would assume this change would be the more correct answer.

rhelmot (approved):

PR #236, opened 2021-09-03
propagate anno

wuruoyu:
It is to solve this: https://github.com/angr/angr/issues/2867#issuecomment-901521771. You can check the issue for more details.

### DIFF
--- a/claripy/ast/base.py
+++ b/claripy/ast/base.py
@@ -1137,7 +1137,8 @@ def simplify(e):
         s._uninitialized = e.uninitialized
         s._uc_alloc_depth = e._uc_alloc_depth
         s._simplified = Base.FULL_SIMPLIFY
-        s.annotations = e.annotations
+        for anno in e.annotations:
+            s = s.annotate(anno)
 
         return s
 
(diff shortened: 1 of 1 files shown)


rhelmot:
I worry that this will cause problems because of hash consing. Annotations are supposed to be part of the AST's identity, but mutating an already-built AST will mutate other references to the same instance. What you want is probably to use the .annotate() function.

wuruoyu:
Thanks for your quick reply. 
Are you suggesting, instead of ```s.annotations = e.annotations```, the following is preferred? 

```
for  anno in e.annotations:
  s.annotate(anno)
```

rhelmot:
Not quite - the point is that once an AST has exited `Base.__new__`, it should never have its identity mutated, so as a result `s.annotate()` is a pure function. You want `s = s.annotate(anno)`.

wuruoyu:
Thanks for your explanation! I have updated the PR accordingly.

rhelmot:
You actually want `s.annotate(*e.annotations)`. The function takes variadic args.

PR #239, opened 2021-09-17
add eq simplification 1^==0

Bonnie1256:


rhelmot:
Can you provide a testcase for this? It's difficult for me to tell what kind of expressions are covered by this.

rhelmot:
@zwimer can you be in charge of verifying that the simplifications being performed here are correct and tested? I tried looking a few times and couldn't get my head around it.

zwimer:
I have a small backlog of tasks but once those are done I can do so.

claripy/simplifications.py
```diff
@@ -195,6 +197,22 @@ def eq_simplifier(a, b):
         if a.op == 'BVV' and b.op != 'BVV':
             return b == a
 
+        # expr - c == 0    ->   expr == c
```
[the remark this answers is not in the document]
zwimer: Maybe length check `a.args` before using `a.args[1]`.

claripy/simplifications.py
```diff
@@ -195,6 +197,22 @@ def eq_simplifier(a, b):
...
+        if a.op == "__sub__" and a.args[1].op == "BVV" and b.op == "BVV":
+            return a.args[0] == a.args[1] + b
+
+        # 1 ^ expr == 0     ->   expr == 1
+        if a.op == '__xor__' and isinstance(b.args[0], int) and b.args[0] == 0:
+            if a.args[0].op == 'If' and a.args[1].args[0] == 1:
+                return a.args[0] == 1
```
zwimer, line 207: These 3 lines read to me as transform:
`( If[v1, v2, v3] ^ op1[1, ...] ^ ...) == b.op[0, ...]` into `If[v1, v2, v3] == 1` 
Where `op1` and `b.op` are unknown ops and where `...` indicates that there may or may not be more arguments, it is not restricted. Something seems wrong here.

zwimer: The following 2 lines have a similar interpretation

claripy/simplifications.py
```diff
@@ -195,6 +197,22 @@ def eq_simplifier(a, b):
...
+            elif a.args[1].op == 'If' and a.args[0].args[0] == 1:
+                return a.args[1] == 1
+
+        if a.op == 'Concat' and isinstance(b.args[0], int) and b.args[0] == 0 and \
+                a.args[0].args[0] == 0 and a.args[1].op == '__invert__':
+            return a.args[1].args[0] == 1
```
zwimer, line 213: The above 3 lines to me read as transform: `Concat[ op1[0, ...], __invert__[x, ...], ... ] == b.op[0, ...]` into `x == 1`. Where `op1` and `b.op` are unknown ops and where `...` indicates that there may or may not be more arguments, it is not restricted. This seems wrong

claripy/simplifications.py
```diff
@@ -280,6 +298,13 @@ def ne_simplifier(a, b):
             # elif b._claripy.is_true(b.args[1] == a) and b._claripy.is_true(b.args[2] == a):
             #	  return b._claripy.false
 
+        # 1 ^ expr != 0     ->   expr == 0
+        if a.op == '__xor__' and b.args[1] == 1 and b.args[0] == 0:
+            if a.args[1].op == "BVV" and a.args[1].args[0] == 1 and a.args[0].size() == 1:
+                return a.args[0] == 0
```
zwimer, line 304: The above 3 lines read to me as transform:
`( x ^ BVV[1, ...] ^ ... ) == b.op[0, 1, ...]` into `x == 0` where `x.size() = 1`. This seems wrong.

zwimer: The following two lines read similarly

zwimer: I stopped reviewing here.

zwimer (changes requested):

claripy/simplifications.py
```diff
@@ -195,6 +197,19 @@ def eq_simplifier(a, b):
...
             return b == a
 
+        # expr - c1 == c2    ->   expr == c1 + c2
+        if a.op == "__sub__" and len(a.args) == 2 and a.args[1].op == "BVV" and b.op == "BVV":
+            return a.args[0] == a.args[1] + b
+
+        # 1 ^ expr == 0     ->   expr == 1
+        if a.op == '__xor__' and b.op == "BVV" and b.args[0] == 0 and len(a.args) == 2:
+            if a.args[0].op == 'If' and a.args[1].op == "BVV" and a.args[1].args[0] == 1:   # If(expr, v1, v2) ^ 1 == 0
+                return a.args[0] == 1
```
zwimer, line 207: Why does the op being `If` matter? 

Also, the comment on 204 doesn't mathematically make sense; if `expr = 1` then `1^1 = 0`, which is invalid. In fact, `1^expr == 0` should always be `False`. The return statement on 207 therefore does not make sense to me.

Also, the comments on lines 204 and 206 conflict; the second seems right, and the comment on line 204 seems to conflict with what the code on 206 does.

claripy/simplifications.py
zwimer: The comment here does not match what this code is doing.

zwimer: Is it meant to be `expr ^ 1 == b -> expr == b` ? If this is the case, I'm not sure why `b.op` matters nor why `a.args[0]` matters.

zwimer: I again stopped reviewing here.

claripy/simplifications.py
zwimer: Oh, wait. `^` is XOR, not power.

zwimer: Though whether we want this is up to @ltfish.

claripy/simplifications.py
```diff
@@ -280,6 +295,13 @@ def ne_simplifier(a, b):
             # elif b._claripy.is_true(b.args[1] == a) and b._claripy.is_true(b.args[2] == a):
             #	  return b._claripy.false
 
+        # 1 ^ expr != 0     ->   expr == 0
+        if a.op == '__xor__' and b.op == "BVV" and b.args[0] == 0 and len(a.args) == 2:
+            if a.args[1].op == "BVV" and a.args[1].args[0] == 1 and a.args[0].size() == 1:
+                return a.args[0] == 0
```
zwimer, line 301: Delete `and a.args[0].size() == 1` and replace `return a.args[0] == 0` with `return a.args[0] != 1` I think.

Unless you intend for this to only work for BVs of length 1-bit (functionally booleans). The logic is correct if that is desired, ping @ltfish for if we want to include this simplifier or if it is too specific or not.

claripy/simplifications.py
```diff
@@ -1000,22 +1063,27 @@ def zeroext_comparing_against_simplifier(op, a, b):
         If the high bits of b are all zeros (in case of __eq__) or have at least one ones (in case of __ne__), ZeroExt
```
zwimer, line 1063: Update the comment to mention op and >=

ltfish (approved):

ltfish:
I'm going to close this PR, and then open a new one after rebasing the changes on the latest claripy master.

claripy/simplifications.py
ltfish: Addressed.

claripy/simplifications.py
ltfish: Good catch. Addressed.

claripy/simplifications.py
ltfish: Updated.

ltfish:
Superseded by #347.

PR #242, opened 2021-10-12
Z3: Use assumptions instead of push/pop

SolalPirelli:
This can be faster since it's a more restricted form of incrementalism (e.g. see https://stackoverflow.com/a/9132505).
I cannot share my private benchmarks (sorry), but in my solver-heavy workload this leads to 5-30% reductions in execution time.

It's also just cleaner code.

### DIFF
--- a/claripy/backends/backend_z3.py
+++ b/claripy/backends/backend_z3.py
@@ -751,21 +751,14 @@ class BackendZ3(Backend):
         global solve_count
 
         solve_count += 1
-        if len(extra_constraints) > 0:
-            solver.push()
-            solver.add(*extra_constraints)
 
-        try:
-            l.debug("Doing a check! (satisfiable)")
-            if solver.check() != z3.sat:
-                return False
-            if model_callback is not None:
-                model_callback(self._generic_model(solver.model()))
-            return True
-        finally:
-            if len(extra_constraints) > 0:
-                solver.pop()
-        return False
+        l.debug("Doing a check! (satisfiable)")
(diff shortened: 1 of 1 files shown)


rhelmot:
> I guess the lint issue on the first commit is because I deleted some "clean" code

Yep, this is by design. We would never reach 100% lint cleanliness otherwise.

> No idea why only 1 of the testing workers failed though. Are they running different tests?

Yes, they are. We have an enormous number of tests so we use our worker pool for parallelism rather than platform diversity.

Unfortunately, it looks like the offending test (angrop/tests/test_rop.py:test_rop_i386_cgc) is not a spurious failure - on master, the test completes without issue but on your branch it takes a very long time and throws multiple timeout errors before failing with an incorrect result. Seems like it could be an edge case with your code? Let me know if you're having issues reproducing.

SolalPirelli:
I did some basic debugging of that ROP test and noticed batch_eval was being called with n=255. I hadn't realized such high values of n were a normal use case, it does make sense that push/pop could be faster if there are a lot of extra assumptions.

I reverted that part of the change: push/pop is still used for batch eval. In theory there is probably some optimal threshold on the value of n below which it makes sense to not use push/pop for perf, but in practice that probably changes with each version of Z3 anyway.

The remaining change keeps most of the perf benefits and the code cleanliness benefits, and passes CI.

I had also replaced `solver.model()` with the existing `model` variable in the 3rd commit, but somehow that seems to have caused a test to fail so I reverted it in the 4th commit.

PR #244, opened 2021-10-14
Correctly create single-precision floating-point values

dnivra:
By default, Python treats floating point numbers as double-precision floats. This becomes a problem when creating concrete single-precision floating point values since value stored would still be double-precision. To work around this, we pack the number to bytes, unpack those bytes and create an FPV from the resulting floating point value.

### DIFF
--- a/claripy/ast/fp.py
+++ b/claripy/ast/fp.py
@@ -1,5 +1,8 @@
+import struct
+
 from .bits import Bits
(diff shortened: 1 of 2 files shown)


rhelmot:
I still consider this a stopgap until we can have fpv.args[0] be a non-double type, but this is good for now.

PR #248, opened 2021-10-31
Z3: clone the solver instead of adding constraints to a fresh instance

CanCebeci:
This results in increased incrementalism and thus speeds up further queries (See https://github.com/Z3Prover/z3/issues/556). Leads to 10-25% reduction in execution time in my private benchmarks.

### DIFF
--- a/claripy/backends/backend_z3.py
+++ b/claripy/backends/backend_z3.py
@@ -701,6 +701,12 @@ class BackendZ3(Backend):
                 s.set('timeout', timeout)
         return s
 
+    def clone_solver(self, s):
+        # This clones the solver.
+        # See https://github.com/Z3Prover/z3/issues/556
+        clone = s.translate(self._context)
+        return clone
+
     def _add(self, s, c, track=False):
         if track:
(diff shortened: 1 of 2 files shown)


rhelmot:
Can you do some tests to check the performance impact on our existing angr/tests/perf tests (should be minimal) and also on some public solving-heavy tests - I would do test_tracer and then the wyvern example in angr-doc

rhelmot:
Also, if you have a workload which you feel like isn't represented by any of our standard perf tests, you can write one and submit a PR and we'll run it every night to check for regressions!

CanCebeci:
I'd gladly do some more performance benchmarks. But before that could you help me understand why that one test fails? I don't understands what it is doing.

rhelmot:
It's a spurious failure. I've run it again and it's passed!

ltfish:
OMG for five years we did not see the second comment in the original Z3 issue about `translate()`...

rhelmot:
I'm going to merge this and fix any problems that come of it :)

PR #249, opened 2021-11-01
Avoid single-argument nodes in simplifications

SolalPirelli:
Having nodes with a single argument, e.g. "add(42)" is rather annoying as they look the same but don't behave the same (e.g. for `structurally_match`) as their lone arg.


rhelmot:
This looks great. Thanks!

PR #251, opened 2021-11-22
Clear caches after simplifying constraints in Z3 backend

dnivra:
During simplification, many Z3 ASTs(and other data) are cached to speed up the process. However, these ASTs references are never freed and thus Z3 treats them as memory leaks. This PR fixes this by clearing the caches after each simplification performed by the Z3 backend.

### DIFF
--- a/claripy/backends/backend_z3.py
+++ b/claripy/backends/backend_z3.py
@@ -974,7 +974,7 @@ class BackendZ3(Backend):
         #l.debug("... after:\n%s", z3_expr_to_smt2(s))
 
         o = self._abstract(s)
-        self.downsize()
+        self._ast_cache.clear()
         o._simplified = Base.FULL_SIMPLIFY
 
         if self._enable_simplification_cache:
--- a/claripy/backends/backend_z3.py
+++ b/claripy/backends/backend_z3.py
@@ -974,6 +974,7 @@ class BackendZ3(Backend):
         #l.debug("... after:\n%s", z3_expr_to_smt2(s))
 
         o = self._abstract(s)
+        self.downsize()
         o._simplified = Base.FULL_SIMPLIFY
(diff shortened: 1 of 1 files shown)


dnivra:
I am not sure why these test cases fail: I am unable to reproduce any of the failures in a local CI setup. I think most of these tests also failed in [CI run on the branch](https://dev.azure.com/angr/claripy/_build/results?buildId=10626) as well. I found that Z3 raises the "invalid dec_ref" error if the reference is reference count is 0, possibly indicating that the reference was never initialized or has been destroyed already. However, from what I found, the AST is cached only in one location in Z3 backend and the reference is increased immediately after caching. Thus, I am unsure how an AST object was cached without a reference count being increased. Is the LRU cache somehow shared; perhaps two tests end up running in the same thread and that causes problems?

ltfish:
I want to better understand the implications before merging in this PR.

zwimer:
The new updates only clears the abstraction cache as I see; I suspect this will be harmful if we abstract similar ASTs from Z3 multiple times, or for example if we extract `A`, `B`, `C`, then `if (A) B else C`. My recommendation to solving the memory issue here is one of these:
1. Expose a toggle in the frontend to clear some / all of these. I.e. make downsize useable. Actually, we already *should* have this: https://github.com/angr/claripy/blob/2e85829f77df3b2823e2aad5c23a3185e27d5851/claripy/frontend.py#L210
2. Start a threading.Thread which periodically checks the memory pressure and calls downsize as needed
3. Have some other method of triggering downsize when needed.

I believe in normal use it is not frequently needed and it primarily clears cache data, so I don't know that auto-clearing it every abstraction is the way to go.

zwimer:
@ltfish Ping

PR #255, opened 2022-01-04
Allow usages of BV to be typechecked

fmagin:
The delayed import construct completely confuses mypy, so this hides it
from the type checker. To allow mypy to recognize that claripy has basic
typing support the `py.typed` file is also added to the `setup.py`


rhelmot:
NICE

PR #259, opened 2022-03-04
Removed generators in test_merging.py

mohitrpatil:


rhelmot (approved):

tests/test_expression.py
```diff
@@ -1,536 +1,533 @@
-import claripy
+#pylint: disable=no-self-use
```
twizmwazin, line 1: Please put a space between the # and pylint. Additionally, these can be comma-seperated lists rather than multiple lines.

tests/test_expression.py
```diff
@@ -1,536 +1,533 @@
...
+#pylint: disable=comparison-with-itself
```
twizmwazin, line 2: This lint should be fixed, not ignored.

tests/test_solver.py
```diff
@@ -1,641 +1,694 @@
-import claripy
-import nose
-import math
+#pylint: disable=no-self-use
+#pylint: disable=line-too-long
```
twizmwazin, line 2: Please fix this rather than ignore it

tests/test_solver.py
```diff
@@ -1,641 +1,694 @@
...
+#pylint: disable=missing-class-docstring
+#pylint: disable=raise-missing-from
```
twizmwazin, line 4: Please fix this one too

tests/test_solver.py
```diff
@@ -1,641 +1,694 @@
...
+
 l = logging.getLogger('claripy.test.solver')
 
 solver_list = (claripy.Solver, claripy.SolverReplacement, claripy.SolverHybrid, claripy.SolverComposite, claripy.SolverCacheless)
 
-def test_solver():
-    for s in solver_list:
-        yield raw_solver, s, True
-        yield raw_solver, s, False
+class TestSolver(unittest.TestCase):
```
twizmwazin, line 15: The test method names in this class can be improved. They're kinda redundant and not normalized to snake case. For example, `test_solver_solver_SolverHybrid` could probably become `test_solver_hybrid`.

tests/test_solver.py
```diff
@@ -1,641 +1,694 @@
...
+#pylint: disable=raise-missing-from
 
+import unittest
 import logging
+import nose
+import claripy
+
 l = logging.getLogger('claripy.test.solver')
 
 solver_list = (claripy.Solver, claripy.SolverReplacement, claripy.SolverHybrid, claripy.SolverComposite, claripy.SolverCacheless)
```
twizmwazin, line 13: This is unused now, yes?

tests/test_solver.py
```diff
@@ -1,641 +1,694 @@
...
+
+        s2.add(z == 12)
+        assert not s2.satisfiable()
+        assert not s2.satisfiable()
+
+    def test_exhaustion(self):
+        s = claripy.Solver()
+        x = claripy.BVS('x', 32)
+        s.add(x >= 19)
+        print(s.min(x, extra_constraints=[x >= 20]))
```
twizmwazin, line 684: While we're here, these print statements could be removed.

twizmwazin (changes requested):

twizmwazin:
As a general rule, I think ignoring `no-self-use` and `missing-class-docstring` are fine in test files, but other lint errors should be fixed rather than ignored.

tests/test_solver.py
mohitrpatil: I removed it.

PR #268, opened 2022-04-22
Use a binary shift instead of exponentiation in `claripy.Extract`

### LINKED ISSUE #267: `claripy.Extract` uses exponentation instead of a binary shift
While profiling some code that used some rather large `claripy.BVV`s, I found that a lot of time was spent in `claripy.Extract`. I found that the main culprit is that the use of exponentiation to construct a bitmask instead of a shift. I used the below script to check that changing this exponentiation by a shift can result in a big increase in performance when the parameter values to `claripy.Extract` are large.

#### Current code
https://github.com/angr/claripy/blob/d8eeba64bab32c62980a2ef3a4b0845a227c720b/claripy/bv.py#L290-L291

#### Proposed change
```python
def Extract(f, t, o):
    return BVV((o.value >> t) & ((1 << (f + 1)) - 1), f + 1 - t)
```

#### Test
`main.py`:
```python
import claripy

def main():
    data = claripy.BVV(0, 25165824)
    for _ in range(100):
        claripy.Extract(25165823, 25133056, data)

if __name__ == "__main__":
    main()
```
With the current implementation, running this script takes 7.377 seconds using the command `time python main.py`. With the proposed improvement, running the same command takes only 0.480 seconds.

LukeSerne:
Closes #267

### DIFF
--- a/claripy/bv.py
+++ b/claripy/bv.py
@@ -288,7 +288,7 @@ def SignExt(num, o):
     return BVV(o.signed, o.bits + num)
 
 def Extract(f, t, o):
-    return BVV((o.value >> t) & (2**(f+1) - 1), f-t+1)
+    return BVV((o.value >> t) & ((1 << (f + 1)) - 1), f + 1 - t)
 
 def Concat(*args):
     total_bits = 0


ltfish:
Thank you!

PR #269, opened 2022-04-22
Make the mask `claripy.Extract` uses smaller to avoid double truncate

LukeSerne:
`claripy.Extract` currently uses a mask of `f` 1-bits, and the resulting value is put in a `BVV` with only `f - t + 1` bits. The result is the same if only `f - t + 1` bits were used in the initial mask, so this PR makes the mask just as big as it should be to avoid another truncation when the `BVV` is constructed. This improves the performance if both `f` and `t` are large.


ltfish:
Thanks!

PR #275, opened 2022-06-19
remove dependency future

FantasqueX:
As angr requires a minimal Python version 3.8 and there is no use of __future__ and builtins, I think it's safe to remove future from dependencies.


rhelmot:
thanks!

PR #277, opened 2022-07-01
a hack to fix state merging tests

Kyle-Kyle:


rhelmot:
Can you explain this change?

Kyle-Kyle:
I have no idea what this patch is. But the veritesting test crashes here. When angr tries to simply `[<SAO <Bool BoolS(cond_0_0_-1)>>, <SAO <Bool BoolS(cond_0_0_-1)>>], [<SAO <Bool BoolS(cond_1_1_-1)>>, <SAO <Bool !BoolS(cond_0_0_-1)>>]`, somehow the argument to `_handle_annotations` will be `'cond_0_110_-1'`, which is a `str` and then crashes.

This patch is a hack to avoid that situation. But this definitely does not fix the root cause of the crash.

PR #283, opened 2022-07-18
Handle sigint explicitly by interrupting all z3 contexts

rhelmot:


claripy/backends/backend_z3.py
```diff
@@ -17,6 +18,26 @@
...
+old_handler = None
+def handle_sigint(signals, frametype):
+    if old_handler == signal.SIG_IGN:
+        return
+
+    contexts = list(ALL_Z3_CONTEXTS)
+    for context in contexts:
+        context.interrupt()
+
+    if old_handler is signal.default_int_handler or old_handler is signal.SIG_DFL:
```
zwimer, line 31: Why does old being `SIG_DFL` lead to raising?

PR #287, opened 2022-08-22
pin to z3 == 4.10.2.0 since 4.11 seems to have introduced serious performance regressions

rhelmot:
I think this is the cause of CI suddenly falling over as noticed in https://github.com/angr/cle/pull/341


rhelmot:
https://github.com/Z3Prover/z3/issues/6294

Issue #292, opened 2022-09-29
`Bits`' `__init__` changes `.length` after hashes are calculated


zwimer:
When creating a `Bits` AST object, the `__init__` function sets `self.length`; in some cases this does not change anything, but in some cases the newly set length is different from the old length.

This occurs after `Base`'s `__new__` runs, and thus it might change the internals of a base, after `Base` calculates the hash of the object to store it inside the lookup cache.

Note that `.length` is a `HASHCONS` variable, it affects the hash.

zwimer:
Actually, this looks like an interaction with `strings.py`'s `__init__`.

Note that here `strings` edits `kwargs`:
https://github.com/angr/claripy/blob/696f1c08c2996c5350867685462d1245070a996d/claripy/ast/strings.py#L31

`Bits`' `self.length = kwargs['length']` would be a no-op if `kwargs` were not edited between `Bits` `__init__` and `Base` `__new__`.

zwimer:
Suggested solution:
1. Refactor all String ops not to use bit lengths rather than byte lengths.
2. Remove `kwargs['length'] *= 8` from `String`'s `__init__` (really this is part of 1)
3. Remove `self.length = kwargs['length']` from `Bits` `__init__`. If this is done without doing 1 & 2 then string support will break

zwimer:
This can be triggered from within `Base`'s new function via the concrete backend constructing an object, it being returned from new, then edited after `__init__` runs.

zwimer:
Note: fixing it might require refactoring `Base`'s `__new__` method as well. Specifically, `Base` can call the concrete backend internally to construct an object. But `Backend.call` only takes in `op` and `args`. For some ops, `length` is not stored in either. For example, a `StringV` might have args `('12', 2)`, but length `80000`- such as the result of an simplified `IntToStr` operation (you can find this is `test_backend_smt.py`).

Since `Backend.call` cannot accept args beyond `op` and `args`, the `StringV` to be constructed will have the size of `.args[1] * 8` (the bit-length of the python `str` held inside the `StringV`) rather than the `StringV` length of `.length`.

Suggested fix: Make `StringV` and any other ops that fall into this category hold both lengths or only the `.length` length. Alternatively `call` could take in `kwargs`, but that feels like it is inviting abuse.

zwimer:
Proposed fixes:
[...]

PR #293, opened 2022-10-04
Change Bits and String not to edit `ast.length`

zwimer:
Fix for: https://github.com/angr/claripy/issues/292

Warning: This changes the semantic meaning of `String.length` from byte length to bit length; use `.string_length` for byte length.

Also, `.size()` and `__len__` are both defined by `Bits`, they also change from byte to bit length; we can override these functions in the `String` class if we desire before merging this.

### DIFF
--- a/claripy/ast/base.py
+++ b/claripy/ast/base.py
@@ -110,7 +110,7 @@ class Base:
         :param args:                The arguments to the AST operation (i.e., the objects to add)
         :param variables:           The symbolic variables present in the AST (default: empty set)
         :param symbolic:            A flag saying whether or not the AST is symbolic (default: False)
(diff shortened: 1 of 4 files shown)


zwimer:
The `test_java.test_fauxware` test is failing; which, since it is java, I assume is because of the breaking semantic change.

zwimer:
Fixed the string length calculations. 

The only failing test case is `test_techniques.test_indirectcfi`; this failed due to a timeout.

@ltfish I reran the test and it passed. I'm not sure if this is an unstable test or if this indicates an issue.

zwimer:
Ping @ltfish

ltfish (approved):

Issue #271, opened 2022-05-16
Exception when a possibly negative value is cached for a shift amount


SolalPirelli:
Run the following:

```python
import angr
import claripy

proj = angr.Project("/bin/true")
state = proj.factory.blank_state()

x = claripy.BVS("x", 64)
state.solver.add(x < 100)

# Comment out this assert and the problem goes away (presumably because it's in the concrete backend used as a cache)
assert state.solver.satisfiable()

problem = 1 >> (1 - x)
state.solver.add(problem >= 0)
assert state.solver.satisfiable()
```

This leads to `claripy.errors.BackendError: <claripy.backends.backend_concrete.BackendConcrete object at 0x7eff53c67880> can't handle operation __rshift__ (BV) due to a failed conversion on a child node`.

Turn the `>>` into a `claripy.LShR` and the error message is more informative:

```
...
  File "/home/solal/.virtualenvs/test/lib/python3.8/site-packages/claripy/bv.py", line 444, in LShR
    return BVV(a.value >> b.signed, a.bits)
ValueError: negative shift count
```

This can be fixed by replacing `.signed` by `.value` in the right places in `bv.py` but since `__rshift__` has a comment `arithmetic shift uses the signed version`, that does not sound right. Perhaps the code should check for a negative signed value and return 0 in that case?

rhelmot:
What does z3 give you when you give it a similar problem?

PR #297, opened 2022-10-18
expr: fix shift with negative operand

tharvik:
related to #271 

when shifting a BV by a negative amount (as might happen when symbolically executing it), it throws a `ValueError: negative shift count` because the count is taken as signed. that should not be the case and this PR fixes that.

also added a bunch of test for the various shift behavior.

btw, I didn't keep the `o.signed < self.bits` branch, as the value is already sanitized in `__init__`, but maybe I missed smth?


rhelmot:
Yes, I believe this is a good change! Thank you!

PR #300, opened 2022-10-27
Condense min/max into singular function

zwimer:
This PR:
1. Condenses min/max into a single function body to avoid code duplication
2. Fixes the bug where `_max` uses `GT` instead of `GE` like min does.
3. Only keeps the latest binary search constraint during the search rather than every previously valid constraint; this should greatly decrease the number of constraints passed to Z3.

### DIFF
--- a/claripy/backends/backend_z3.py
+++ b/claripy/backends/backend_z3.py
@@ -852,8 +852,10 @@ class BackendZ3(Backend):
 
         return result_values
 
-    @condom
-    def _min(self, expr, extra_constraints=(), signed=False, solver=None, model_callback=None):
+    def _extrema(self, is_max: bool, expr, extra_constraints, signed, solver, model_callback):
+        """
+        _max if is_max else _min
+        """
         global solve_count
 
         if signed:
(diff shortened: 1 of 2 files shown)


zwimer:
Note that there is no need to test `if hi == lo`, since we looped until `hi - lo > 1`; thus the previous iteration `hi - lo > 1`, since `hi`, middle, and `lo` must be integers, this implies `hi - lo >= 2` thus the previous `middle` would not equal either, and thus the new `hi` and `lo` would not be equal, since only one is now what the previous `middle` was.

zwimer:
The latest commit adds what should be a pretty big optimization to min/max; we don't need a list of historically valid constraints while binary searching, we only need the latest constraint to determine if 'this half' is the 'good half' or if we select the 'other half' next.

zwimer:
I ran the `test_z3.py` script on master vs this branch, I found a speed up from 9.5 seconds down to 7.5 on average. @ltfish @zardus

zwimer:
That is by no means a comprehensive test, and it is only on `8`-bit BVs, just thought it would be worth mentioning this comes with a performance boost.

rhelmot (approved):

PR #302, opened 2022-10-27
Fix: Update z3 usage to use actual z3 functions

zwimer:
The old function did not exist in Z3


ltfish:
1. Test cases fail because of https://github.com/angr/angr-management/pull/760.
2. Looks like we have no test cases covering the line that you just fixed?

zwimer:
1. Correct
2. Correct

Issue #303, opened 2022-10-27
pytest skips tests


zwimer:
### Description

When running `pytest` on claripy, it skips certain test directories because: `PytestCollectionWarning: yield tests were removed in pytest 4.0 - test_composite_discrepancy will be ignored`

### Steps to reproduce the bug

```bash
git clone git@github.com:angr/claripy
cd claripy
pytest 
```

### Environment

A clean install of `claripy` master, this commit: https://github.com/angr/claripy/commit/0c1f0e2c6da5655216f032b304adff66eae190a5

I created a new virtualenv and installed `pytest` and `pytest-xdist`.

This was on an M1 Mac.

### Additional context

Actual output:
```
(test) zwimer@Lotus ~/D/real_c> pytest -n9                                                                                                                                                                                                lint-bk-z3
================================================================================================================ test session starts ================================================================================================================
platform darwin -- Python 3.10.8, pytest-7.1.3, pluggy-1.0.0
rootdir: /Users/zwimer/Desktop/real_c
plugins: xdist-3.0.2
gw0 [255] / gw1 [255] / gw2 [255] / gw3 [255] / gw4 [255] / gw5 [255] / gw6 [255] / gw7 [255] / gw8 [255]
.................................................sssss......sssssssss.......s.sssss..s.ssss.s...................ss..ssssss.....ssssss.sss..sssss.sssss..s.s..ss..ss....s.........................................xxxxx....xx.x......x....x... [ 92%]
..................                                                                                                                                                                                                                            [100%]
================================================================================================================= warnings summary ==================================================================================================================
tests/test_solver.py:17
tests/test_solver.py:17
tests/test_solver.py:17
tests/test_solver.py:17
tests/test_solver.py:17
tests/test_solver.py:17
tests/test_solver.py:17
tests/test_solver.py:17
tests/test_solver.py:17
  tests/test_solver.py:17: PytestCollectionWarning: yield tests were removed in pytest 4.0 - test_solver will be ignored
    def test_solver():

tests/test_solver.py:23
tests/test_solver.py:23
tests/test_solver.py:23
tests/test_solver.py:23
tests/test_solver.py:23
tests/test_solver.py:23
tests/test_solver.py:23
tests/test_solver.py:23
tests/test_solver.py:23
  tests/test_solver.py:23: PytestCollectionWarning: yield tests were removed in pytest 4.0 - test_hybrid_solver will be ignored
    def test_hybrid_solver():

tests/test_solver.py:77
[...]

PR #307, opened 2022-10-28
Edit test_solver so tests actually run

zwimer:
Fixes: https://github.com/angr/claripy/issues/303

### DIFF
--- a/tests/test_solver.py
+++ b/tests/test_solver.py
@@ -1,28 +1,10 @@
 from common_backend_smt_solver import if_installed
+from unittest import TestCase, main
 import claripy
 
 import logging
 
 l = logging.getLogger("claripy.test.solver")
-
-solver_list = (
-    claripy.Solver,
-    claripy.SolverReplacement,
-    claripy.SolverHybrid,
-    claripy.SolverComposite,
-    claripy.SolverCacheless,
-)
-
-
-def test_solver():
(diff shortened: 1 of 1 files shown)


zwimer:
Also leads to running more tests in `__main__` that were previously not run.

twizmwazin:
Why was this merged with a failing lint? @zwimer can you make a follow up that addresses the lint issues?

zwimer:
@twizmwazin Most all of the issues in the lint are either `Method could be a function (no-self-use)`, though unittest examples all take in `self` so assuming we want to follow that convention I'd ignore each issue there. The other big issue is `Instance of 'Base' has no 'solver' member (no-member)`; but that is intentional, it if something tries to run these without defining `solver` in a subclass it should error, it shouldn't start executing as it might if we had `solver=None`; so I'd also ignore that warning.

If I add a pylint disable these to the top of the file, will that disable them for the entire file? Otherwise we'd just have to spam disable warnings everywhere which I am ok with not-doing on a test case file

twizmwazin:
I agree with no-self-use and docstrings being superfluous here, but wrong-import-order, no-member, duplicate entry in set all seem like things that should be resolved. no-member could be better solved with an abstract class and still achieve the behavior you want. Adding a pylint ignore at the top of a file or class applies to the entire file/class.

PR #313, opened 2022-11-03
Only edit signal handling in main thread

zwimer:


rhelmot (changes requested):
I believe this will not work - old_handler must be initialized to _something_, and this will leave it undefined if claripy is imported in a non-main thread.

zwimer:
@rhelmot Correct me if I'm wrong, but `old_handler` should only ever be used in the signal handler after defined, which should only be invoked in the main thread.

rhelmot:
Oops! Yes, that is correct.

zwimer:
I can revert the last commit. I believe it might just add a few cycles to import via the syscall; but it doesn't harm anything. Lmk which you prefer

rhelmot (approved):

rhelmot:
I don't see how either version will cause problems now, but I approve regardless.

PR #319, opened 2022-11-05
String symbols abstraction explicitly not supported by z3 backend

zwimer:
Not supported because z3 does not supported strings
https://github.com/Z3Prover/z3/discussions/6443


zwimer:
No point in CI right now; this will fail as the length was not set during string abstraction (a syntax error)

zwimer:
This may not be possible given: https://github.com/Z3Prover/z3/discussions/6443#discussioncomment-4081446

zwimer:
Ping @ltfish

zwimer:
I'm going to merge this; if it causes issues we can always revert; though if this causes issues, that probably indicates the thing with an issue is broken, not this.

PR #320, opened 2022-11-06
CompositeSolver: The merged child solver should be of the transitive closure of variables.

ltfish:
Close https://github.com/angr/angr/issues/3604.


ltfish:
It's a bug! Thanks for catching it.

PR #333, opened 2023-01-23
Enable ruff in pre-commit

twizmwazin:
This one definitely has a not great set of ignore for ruff. Some of these are real problems we should address separately.


zwimer:
@twizmwazin Please resolve merge conflicts.

twizmwazin:
Superseded by #401

Issue #338, opened 2023-02-14
Expressions annotated with SimplificationAvoidanceAnnotation cause an AttributeError during simplification


mounir-khaled:
### Description

Expressions annotated with SimplificationAvoidanceAnnotation cause an AttributeError during simplification

### Steps to reproduce the bug

Running this example:
```
import claripy                                                     
x = claripy.BVS('x', 32)                                           
b = (x == 0).annotate(claripy.SimplificationAvoidanceAnnotation())
claripy.simplify(claripy.Or(False, claripy.And(b, b))) 
```

Raises an exception:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "...\venv\lib\site-packages\claripy\operations.py", line 57, in _op
    simp = _handle_annotations(simplifications.simpleton.simplify(name, fixed_args), args)
  File "...\venv\lib\site-packages\claripy\simplifications.py", line 41, in simplify
    return self._simplifiers[op](*args)
  File "...\venv\lib\site-packages\claripy\simplifications.py", line 378, in boolean_and_simplifier
    if flattened.op != "And":
AttributeError: 'NoneType' object has no attribute 'op'
```

PR #339, opened 2023-02-14
Fix AttributeError in boolean_and_simplifier when _flatten_simplifier returns None

mounir-khaled:
This should fix this issue https://github.com/angr/claripy/issues/338


rhelmot:
ty!

PR #341, opened 2023-02-17
Update z3 to version 4.12.1.0

twizmwazin:


twizmwazin:
The bug I was hoping might be solved with this isn't and the Ci failures are non-trivial enough that I don't plan on finishing this.

PR #345, opened 2023-03-24
Bit length vs Byte length bug fix in string_length

zwimer:
Bit length vs Byte length bug fix in string_length

### DIFF
--- a/claripy/ast/strings.py
+++ b/claripy/ast/strings.py
@@ -26,10 +26,10 @@ class String(Bits):
 
     def __init__(self, *args, length: int, **kwargs):
         """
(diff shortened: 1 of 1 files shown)


zwimer:
@rhelmot In case it is unintended, making an `IntToStr` with args `(<BV32 0xc>,)` and kwargs `{'length': 80000, 'uninitialized': None}` leads to the eager backend logic in `__new__` invoking leading to creating a different AST with a different length.

I.e. The output length is 16 bits long instead of 80000 bits.

I updated the PR to keep `.length` and `.bit_length` sync'd given this, but it might be worth a look if that isn't expected.

PR #354, opened 2023-05-15
Update z3 to 4.12.2.0

twizmwazin:
z3 finally released an update with my patch that fixes pyinstaller compatibility with newer versions. I'd like to sort out any compatibility issues and try to update.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -22,7 +22,7 @@ install_requires =
     cachetools
     decorator
     pysmt>=0.9.5
-    z3-solver==4.10.2.0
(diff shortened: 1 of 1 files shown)


twizmwazin:
This is a much better result than last time, but I'm not really sure how to best debug a timeout here.

syheliel:
the ci is re-triggered. It seems that the upgrade causes some real problems. You can have a look at what the CI does in https://github.com/angr/ci-settings and debug it in your local machine.

rhelmot:
Bestie you're talking to the guy who built ci-settings

PR #399, opened 2024-05-27
Fix double z3 check

salls:
in some cases, such as no constraints in the state, this code was calling z3 twice for a single eval. @zardus you might want to check this is the correct way to fix it


zardus:
Good catch on the empty solver double query. Don't even know how you'd think to notice that :-)

PR #401, opened 2024-06-19
Add ruff linter

twizmwazin:
We never got around to doing this for claripy


twizmwazin:
Last week I was under the impression that the `test_techniques.test_indirectcfi` might have been related to changes here, but with today's pre-commit update the test failed every time. So I'm just going to disable that test for now. It should be fixed but I don't want to be blocked on it.

PR #404, opened 2024-06-20
Refactor into Rust

twizmwazin:
This is an idea I want to experiment with. Basically the idea is that for each class in claripy, make a native class that becomes its base class. Then, one-by-one translate member variables and methods into the base class until the python class is empty. At each step of the way, keep the whole thing integrated and working (hence this PR, I want to test angr in conjunction with my changes). There's going to probably be little or negative performance gains in the early stages of this, but when large sections (ie, the frontend) are fully translated, that's when a speedup might be noticeable.


ltfish:
Excited!

PR #418, opened 2024-07-22
Remove unused backends

twizmwazin:
I think these backends are not being used. In pursuit of modernizing claripy, if angr can work without these backends, let's remove them.

link: https://github.com/angr/angr-examples/pull/8


ltfish:
I am fairly certain that many of these backends were developed as part of the effort of supporting string solvers. @zardus what do you think of aggressively removing claripy components that are not used in angr but may be used in other projects?

twizmwazin:
Regarding string solvers, the z3 backend already appears to be a drop-in replacement for the existing string solving support, without any additional code or dependencies. So this should not be a regression solver-capabilities.

ltfish:
> So this should not be a regression in solver capabilities.

I am asking Lukas and Fabio what the original uses were. They implemented this feature to support multiple string solvers.

Issue #444, opened 2024-08-19
`TestStrings.test_index_of_symbolic_start_idx` usually hangs


twizmwazin:
### Description

This test was originally part of the smtlib solver tests. When I removed the smtlib solver support, I kept the tests around to test the strings functionality in claripy and the z3 backend since they seemed useful, but this test in particular usually hung. This should be debugged and addressed so the test can be re-enabled.

PR #453, opened 2024-08-29
Issue 440 breakdown tests

spshah1701:
Implemented the replace_unicode method to convert Unicode escape sequences to their corresponding characters for proper string handling. Resolved the test hanging issue by making the symbolic constraints more manageable for the solver.


spshah1701:
The issue being resolved in this PR is #444.

twizmwazin:
This test is failing consistently now which is better than inconsistently, we'll need to figure out why z3 is being cancelled.

PR #473, opened 2024-09-07
Add __len__ method to Bool

bluesadi:


ltfish:
I don't think this is the correct fix. I bet you encountered an exception in angr decompiler; the correct fix should be converting Bool into BVV/BVS before it's used as a bitvector in angr.

ltfish:
On a second thought, I think this is fine (because Bool has .size() anyway). It may or may not fix the real issue in angr decompiler though.

ltfish:
Thanks!

PR #491, opened 2024-09-13
add a type annotation

rhelmot:


twizmwazin:
We should probably make length not optional at some point.... but that's a larger problem.

PR #593, opened 2025-02-03
balancer: make iterative instead of recursive

rhelmot:


rhelmot:
Typechecking fails because I added type annotations which fix "critical" errors and unblock a zillion "less-critical" errors. I'm not fixing those.

PR #609, opened 2025-03-16
Remove pow op from BVs

twizmwazin:
This is a test to see if this is actually used anywhere. It isn't supported in any backends, and smtlib nor z3 supports this op.


twizmwazin:
Well I'l take that as confirmation

PR #619, opened 2025-04-11
simplify signal handler as signal.default_int_handler is callable

demoray:
Rather than directly raising KeyboardInterrupt, use signal.default_int_handler as it is Callable.


twizmwazin:
Thanks!

PR #621, opened 2025-04-15
bitwise_and_simplifier: do not crash on non-bit-sized ASTs

rhelmot:
Was triggered around level 4814.


ltfish:
Ah you beat me to it…

ltfish:
Why are we simplifying non-bit-sized ASTs in bitwise-and simplifier?

rhelmot:
`__and__` works on other sizes of AST, check the previous cases in the simplifier.

PR #626, opened 2025-05-14
Add fpToFPUnsigned bound op

twizmwazin:
Fixes fpToFPUnsigned from having a None length when being returned from the z3 backend. Fixes https://github.com/angr/angr/issues/5456


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/claripy_626

PR #698, opened 2026-02-20
Bump pinned z3 version to 4.16.0.0 (latest)

rhelmot:
Secret motivation for this: see if it fixes the damn segfaults in CI


calebh:
@rhelmot is this bump to Z3 only failing on Mac, or is there also an issue on Linux? Do the failing "angr Ecosystem Test" cases indicate a more widespread failure?

rhelmot:
angr ecosystem tests indicate wider issues. I don't recall what the problem was, so let's rebase this and see what happens...

rhelmot:
And now following the API fix, all the failures are segfaults. We have since fixed the segfaults I was complaining about in the original post, so these are all new ones caused by z3 :)

PR #714, opened 2026-04-28
Add bool Xor; improve simplification

rhelmot:
- Add option to simplify during replacements
- add `__xor__` to Bool ASTs
- add some trivial simplifiers to And/Or/Xor
- for commutative ops, sort arguments by stable hash during simplification

These were all moderately helpful during a CTF.


rhelmot:
image

:|

PR #723, opened 2026-05-26
Z3Backend: Fix assertion error when SIGINT handler not installed

tedanvosin:
- Uninstall SIGINT handler only when it was installed.
- Fixes assertion error when running symbolic execution in angr-management.


rhelmot:
fix the rebase on this and I shall merge

tedanvosin:
Fixed

PR #737, opened 2026-07-21
Use z3-solver 5.1.0.0 on every platform

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

### Problem

claripy pins `z3-solver==4.13.0.0`, which has no Emscripten wheel, so claripy cannot be installed under Pyodide at all. That pin also stopped being co-installable with angr, which pins `z3-solver==5.1.0.0` on master.

An earlier revision of this pull request solved the first half with a platform-split marker — 4.13.0.0 natively, Z3 5 on Emscripten. That was rejected: one z3 version everywhere, or nothing.

### Root cause

Three things stood between claripy and a single pin.

**Only one release covers every platform.** `z3-solver==5.1.0.0` is the only release that ships a `pyemscripten_2026_0_wasm32` wheel, alongside manylinux x86_64/aarch64/riscv64, macOS 13 arm64/x86_64 and win32/win_amd64/win_arm64. Checked across all 52 released versions on 2026-09-08. 4.13.0.0 and 5.0.0.0 have no Emscripten wheel.

**The native blow-up has a mitigation.** On 5.1.0.0, angr's `test_manyfloatsum_symbolic_i386` and `test_manyfloatsum_symbolic_x86_64` exhausted memory and ended in `solver unknown: out of memory`. Turning z3's `rewriter.ite_extra_rules` off makes both pass at about 621 MB peak RSS on the same stock wheel, `libz3.so.5.1` byte-identical between the two arms, no z3 rebuilt or patched.

That parameter is a mitigation, not the cause. Its default is `True` in `src/params/bool_rewriter_params.pyg` at both `z3-4.13.0` and `z3-5.1.0` — the file is byte-identical between the two tags — and 4.13.0.0 completes both tests with the rules on. The regression itself was bisected earlier on this pull request to z3 `d66609ea` (4.13.1) and `f41134d1` (4.14.0): https://github.com/angr/claripy/pull/737#issuecomment-5052593249

**The FP sign ABI moved.** Reading `z3/z3core.py` out of one wheel from each of the 18 releases between 4.13.0.0 and 5.1.0.0, `Z3_fpa_get_numeral_sign.argtypes` changes from `POINTER(c_int)` to `POINTER(c_bool)` at **4.15.5.0** and stays there. `ctypes` checks pointer types strictly, so `byref(c_int)` raises `ArgumentError` when claripy reads a floating-point model value back through that call.

### Fix

One unmarked pin, `z3-solver==5.1.0.0`, on every platform. With it:

- `z3.set_param("rewriter.ite_extra_rules", "false")`, beside the `rewriter.hi_fp_unspecified` line claripy has set the same way since 4.8.7.0. This is process-global, and what it costs on non-FP workloads is not measured.
- `ctypes.c_bool` at both `Z3_fpa_get_numeral_sign` call sites, and a test that reads a signed floating-point model value back. No existing test did.

### Testing

Validation: https://github.com/angr/claripy/pull/737#issuecomment-5427519925

session: sharpen

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,11 @@ classifiers = [
 ]
 urls = { Homepage = "https://github.com/angr/claripy" }
 requires-python = ">=3.12"
-dependencies = ["cachetools", "z3-solver==4.13.0.0"]
+dependencies = [
+    "cachetools",
+    "z3-solver==4.13.0.0; sys_platform != 'emscripten'",
+    "z3-solver>=5.0.0.0; sys_platform == 'emscripten'",
+]
 dynamic = ["version"]
 
 [project.readme]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,11 @@ classifiers = [
 ]
 urls = { Homepage = "https://github.com/angr/claripy" }
 requires-python = ">=3.12"
-dependencies = ["cachetools", "z3-solver==5.0.0.0"]
+dependencies = [
+    "cachetools",
+    "z3-solver==4.13.0.0; sys_platform != 'emscripten'",
+    "z3-solver>=5.0.0.0; sys_platform == 'emscripten'",
+]
 dynamic = ["version"]
 
 [project.readme]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,11 +16,7 @@ classifiers = [
 ]
 urls = { Homepage = "https://github.com/angr/claripy" }
 requires-python = ">=3.12"
-dependencies = [
-    "cachetools",
-    "z3-solver==4.13.0.0; sys_platform != 'emscripten'",
-    "z3-solver>=5.0.0.0; sys_platform == 'emscripten'",
-]
+dependencies = ["cachetools", "z3-solver==5.0.0.0"]
 dynamic = ["version"]
 
 [project.readme]
--- a/claripy/backends/backend_z3.py
+++ b/claripy/backends/backend_z3.py
@@ -753,7 +753,7 @@ class BackendZ3(Backend):
             # TODO: do better than this
             fp_mantissa = float(z3.Z3_fpa_get_numeral_significand_string(ctx, ast))
             fp_exp = int(z3.Z3_fpa_get_numeral_exponent_string(ctx, ast, False))
-            fp_sign_c = ctypes.c_int()  # pylint: disable=no-value-for-parameter
+            fp_sign_c = ctypes.c_bool()  # pylint: disable=no-value-for-parameter
             z3.Z3_fpa_get_numeral_sign(ctx, ast, ctypes.byref(fp_sign_c))
             fp_sign = -1 if fp_sign_c.value != 0 else 1
             return fp_sign * fp_mantissa * (2**fp_exp)
@@ -782,7 +782,7 @@ class BackendZ3(Backend):
             # TODO: do better than this
             fp_mantissa = int(z3.Z3_fpa_get_numeral_significand_string(ctx, ast))
             fp_exp = int(z3.Z3_fpa_get_numeral_exponent_string(ctx, ast, True))
-            fp_sign_c = ctypes.c_int()  # pylint: disable=no-value-for-parameter
+            fp_sign_c = ctypes.c_bool()  # pylint: disable=no-value-for-parameter
             z3.Z3_fpa_get_numeral_sign(ctx, ast, ctypes.byref(fp_sign_c))
             fp_sign = 1 if fp_sign_c.value != 0 else 0
         elif op_name == "MinusZero":
--- a/claripy/backends/backend_z3.py
+++ b/claripy/backends/backend_z3.py
@@ -42,7 +42,6 @@ log = logging.getLogger(__name__)
 
 ALL_Z3_CONTEXTS = weakref.WeakSet()
 INT_STRING_CHUNK_SIZE: int | None = None  # will be updated later if we are on CPython 3.11+
-Z3_BOOL_TYPE = ctypes.c_bool if z3.get_version() >= (5, 0, 0, 0) else ctypes.c_int
 
 
 class SigintHandler:
@@ -300,9 +299,10 @@ class BackendZ3(Backend):
 
         # Per-thread Z3 solver
         # This setting is treated as a global setting and is not supposed to be changed during runtime, unless you know
-        # what you are doing.
+        # what you are doing. Reuse also prevents Z3 5 from retaining substantial native memory across short-lived
+        # solvers during symbolic execution.
         if reuse_z3_solver is None:
-            reuse_z3_solver = os.environ.get("REUSE_Z3_SOLVER", "False").lower() in {"1", "true", "yes", "y"}
+            reuse_z3_solver = os.environ.get("REUSE_Z3_SOLVER", "True").lower() in {"1", "true", "yes", "y"}
         self.reuse_z3_solver = reuse_z3_solver
 
         self._ast_cache_size = ast_cache_size
@@ -754,7 +754,7 @@ class BackendZ3(Backend):
             # TODO: do better than this
             fp_mantissa = float(z3.Z3_fpa_get_numeral_significand_string(ctx, ast))
             fp_exp = int(z3.Z3_fpa_get_numeral_exponent_string(ctx, ast, False))
-            fp_sign_c = Z3_BOOL_TYPE()  # pylint: disable=no-value-for-parameter
+            fp_sign_c = ctypes.c_bool()  # pylint: disable=no-value-for-parameter
             z3.Z3_fpa_get_numeral_sign(ctx, ast, ctypes.byref(fp_sign_c))
             fp_sign = -1 if fp_sign_c.value != 0 else 1
             return fp_sign * fp_mantissa * (2**fp_exp)
@@ -783,7 +783,7 @@ class BackendZ3(Backend):
             # TODO: do better than this
             fp_mantissa = int(z3.Z3_fpa_get_numeral_significand_string(ctx, ast))
             fp_exp = int(z3.Z3_fpa_get_numeral_exponent_string(ctx, ast, True))
-            fp_sign_c = Z3_BOOL_TYPE()  # pylint: disable=no-value-for-parameter
+            fp_sign_c = ctypes.c_bool()  # pylint: disable=no-value-for-parameter
             z3.Z3_fpa_get_numeral_sign(ctx, ast, ctypes.byref(fp_sign_c))
             fp_sign = 1 if fp_sign_c.value != 0 else 0
         elif op_name == "MinusZero":
--- a/claripy/backends/backend_z3.py
+++ b/claripy/backends/backend_z3.py
@@ -42,6 +42,7 @@ log = logging.getLogger(__name__)
 
 ALL_Z3_CONTEXTS = weakref.WeakSet()
 INT_STRING_CHUNK_SIZE: int | None = None  # will be updated later if we are on CPython 3.11+
+Z3_BOOL_TYPE = ctypes.c_bool if z3.get_version() >= (5, 0, 0, 0) else ctypes.c_int
 
 
 class SigintHandler:
@@ -753,7 +754,7 @@ class BackendZ3(Backend):
             # TODO: do better than this
             fp_mantissa = float(z3.Z3_fpa_get_numeral_significand_string(ctx, ast))
             fp_exp = int(z3.Z3_fpa_get_numeral_exponent_string(ctx, ast, False))
-            fp_sign_c = ctypes.c_bool()  # pylint: disable=no-value-for-parameter
+            fp_sign_c = Z3_BOOL_TYPE()  # pylint: disable=no-value-for-parameter
             z3.Z3_fpa_get_numeral_sign(ctx, ast, ctypes.byref(fp_sign_c))
             fp_sign = -1 if fp_sign_c.value != 0 else 1
             return fp_sign * fp_mantissa * (2**fp_exp)
@@ -782,7 +783,7 @@ class BackendZ3(Backend):
             # TODO: do better than this
             fp_mantissa = int(z3.Z3_fpa_get_numeral_significand_string(ctx, ast))
             fp_exp = int(z3.Z3_fpa_get_numeral_exponent_string(ctx, ast, True))
-            fp_sign_c = ctypes.c_bool()  # pylint: disable=no-value-for-parameter
+            fp_sign_c = Z3_BOOL_TYPE()  # pylint: disable=no-value-for-parameter
             z3.Z3_fpa_get_numeral_sign(ctx, ast, ctypes.byref(fp_sign_c))
             fp_sign = 1 if fp_sign_c.value != 0 else 0
         elif op_name == "MinusZero":
--- a/claripy/backends/backend_z3.py
+++ b/claripy/backends/backend_z3.py
@@ -42,6 +42,7 @@ log = logging.getLogger(__name__)
 
 ALL_Z3_CONTEXTS = weakref.WeakSet()
 INT_STRING_CHUNK_SIZE: int | None = None  # will be updated later if we are on CPython 3.11+
+Z3_BOOL_TYPE = ctypes.c_bool if z3.get_version() >= (5, 0, 0, 0) else ctypes.c_int
 
 
 class SigintHandler:
@@ -299,10 +300,9 @@ class BackendZ3(Backend):
 
         # Per-thread Z3 solver
         # This setting is treated as a global setting and is not supposed to be changed during runtime, unless you know
-        # what you are doing. Reuse also prevents Z3 5 from retaining substantial native memory across short-lived
-        # solvers during symbolic execution.
+        # what you are doing.
         if reuse_z3_solver is None:
-            reuse_z3_solver = os.environ.get("REUSE_Z3_SOLVER", "True").lower() in {"1", "true", "yes", "y"}
+            reuse_z3_solver = os.environ.get("REUSE_Z3_SOLVER", "False").lower() in {"1", "true", "yes", "y"}
         self.reuse_z3_solver = reuse_z3_solver
 
         self._ast_cache_size = ast_cache_size
@@ -754,7 +754,7 @@ class BackendZ3(Backend):
             # TODO: do better than this
             fp_mantissa = float(z3.Z3_fpa_get_numeral_significand_string(ctx, ast))
             fp_exp = int(z3.Z3_fpa_get_numeral_exponent_string(ctx, ast, False))
-            fp_sign_c = ctypes.c_bool()  # pylint: disable=no-value-for-parameter
+            fp_sign_c = Z3_BOOL_TYPE()  # pylint: disable=no-value-for-parameter
             z3.Z3_fpa_get_numeral_sign(ctx, ast, ctypes.byref(fp_sign_c))
             fp_sign = -1 if fp_sign_c.value != 0 else 1
             return fp_sign * fp_mantissa * (2**fp_exp)
@@ -783,7 +783,7 @@ class BackendZ3(Backend):
             # TODO: do better than this
             fp_mantissa = int(z3.Z3_fpa_get_numeral_significand_string(ctx, ast))
             fp_exp = int(z3.Z3_fpa_get_numeral_exponent_string(ctx, ast, True))
-            fp_sign_c = ctypes.c_bool()  # pylint: disable=no-value-for-parameter
+            fp_sign_c = Z3_BOOL_TYPE()  # pylint: disable=no-value-for-parameter
             z3.Z3_fpa_get_numeral_sign(ctx, ast, ctypes.byref(fp_sign_c))
             fp_sign = 1 if fp_sign_c.value != 0 else 0
         elif op_name == "MinusZero":
--- a/tests/test_fp.py
+++ b/tests/test_fp.py
@@ -59,6 +59,13 @@ class TestFp(unittest.TestCase):
         s = claripy.Solver()
         assert s.eval(b, 1)[0] == 2
 
+    def test_fp_model_value(self):
+        a = claripy.FPS("a", claripy.FSORT_FLOAT)
+        s = claripy.Solver()
+        s.add(a == -2.5)
+        assert s.eval(a, 1) == (-2.5,)
+        assert s.eval(a.raw_to_bv(), 1) == (0xC0200000,)
+
     def test_fp_precision_loss(self):
         dd = claripy.FPV(101817237862.0, claripy.FSORT_DOUBLE)
         s = claripy.Solver()
--- a/tests/test_solver.py
+++ b/tests/test_solver.py
@@ -5,17 +5,6 @@ from unittest import TestCase, main
 
 import claripy
 import claripy.frontend
-from claripy.backends.backend_z3 import BackendZ3
-
-
-def test_z3_solver_reuse_defaults_on(monkeypatch):
-    monkeypatch.delenv("REUSE_Z3_SOLVER", raising=False)
-    assert BackendZ3().reuse_z3_solver
-
-
-def test_z3_solver_reuse_can_be_disabled(monkeypatch):
-    monkeypatch.setenv("REUSE_Z3_SOLVER", "false")
-    assert not BackendZ3().reuse_z3_solver
 
 
 def raw_hybrid_solver(reuse_z3_solver):
--- a/tests/test_solver.py
+++ b/tests/test_solver.py
@@ -5,6 +5,17 @@ from unittest import TestCase, main
 
 import claripy
 import claripy.frontend
+from claripy.backends.backend_z3 import BackendZ3
+
+
+def test_z3_solver_reuse_defaults_on(monkeypatch):
+    monkeypatch.delenv("REUSE_Z3_SOLVER", raising=False)
+    assert BackendZ3().reuse_z3_solver
+
+
+def test_z3_solver_reuse_can_be_disabled(monkeypatch):
+    monkeypatch.setenv("REUSE_Z3_SOLVER", "false")
+    assert not BackendZ3().reuse_z3_solver
 
 
 def raw_hybrid_solver(reuse_z3_solver):
--- a/uv.lock
+++ b/uv.lock
@@ -129,7 +129,8 @@ name = "claripy"
 source = { editable = "." }
 dependencies = [
     { name = "cachetools" },
-    { name = "z3-solver" },
+    { name = "z3-solver", version = "4.13.0.0", source = { registry = "https://pypi.org/simple" }, marker = "sys_platform != 'emscripten'" },
+    { name = "z3-solver", version = "5.0.0.0", source = { registry = "https://pypi.org/simple" }, marker = "sys_platform == 'emscripten'" },
 ]
 
 [package.dev-dependencies]
@@ -147,7 +148,8 @@ docs = [
 [package.metadata]
 requires-dist = [
     { name = "cachetools" },
-    { name = "z3-solver", specifier = "==5.0.0.0" },
+    { name = "z3-solver", marker = "sys_platform != 'emscripten'", specifier = "==4.13.0.0" },
+    { name = "z3-solver", marker = "sys_platform == 'emscripten'", specifier = ">=5.0.0.0" },
 ]
 
 [package.metadata.requires-dev]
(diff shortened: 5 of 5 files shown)


zardus:
Tandem PRs — please test these together:

- https://github.com/angr/vex/pull/82
- https://github.com/angr/pyvex/pull/555
- https://github.com/angr/claripy/pull/737
- https://github.com/angr/cle/pull/702
- https://github.com/angr/cle/pull/703
- https://github.com/angr/cle/pull/704
- https://github.com/angr/pypcode/pull/283
- https://github.com/angr/angr/pull/6658

pyproject.toml
```diff
@@ -16,7 +16,11 @@ classifiers = [
 ]
 urls = { Homepage = "https://github.com/angr/claripy" }
 requires-python = ">=3.12"
-dependencies = ["cachetools", "z3-solver==4.13.0.0"]
+dependencies = [
+    "cachetools",
+    "z3-solver==4.13.0.0; sys_platform != 'emscripten'",
```
twizmwazin, line 21: What if we upgraded to z3 5.0.0.0 across all platforms instead?

zardus: Done in a91c1e78: z3-solver 5.0.0.0 is now used on every platform, and the lockfile has been regenerated. The full Claripy test suite passes locally (331 passed, 1 skipped).

zardus:
Ecosystem CI exposed one Z3 5 binding-type mismatch in floating-point model extraction. d7f1a5be now passes a c_bool output to Z3_fpa_get_numeral_sign (matching the Z3 5 signature) and adds a regression test for signed FP model values. The full Claripy suite passes locally: 332 passed, 1 skipped.

pyproject.toml
zardus: I tried this in a91c1e78, but the native ecosystem runs exposed a severe Z3 5 regression: `angr/tests/procedures/libc/test_scanf.py::TestScanf::test_scanf_multi` peaks at about 15.8 GB RSS and takes ~12 s with Z3 5.0.0.0, versus about 522 MB and ~1.1 s with Z3 4.13.0.0. That OOM-kills the GitHub-hosted runners (and reproduced locally under `/usr/bin/time -v`). faa92806 therefore keeps Z3 5 for Emscripten/Pyodide while retaining 4.13 natively. It also selects the matching `ctypes` boolean representation at runtime, so Claripy supports both Z3 ABIs without any Z3 source changes. The full native Claripy suite passes (332 passed, 1 skipped).

twizmwazin:
If the tests pass, then we should use z3 5.0.0.0 across all platforms. Platform specific version discrepancies is a footgun waiting to be fired.

zardus:
Done in 2c50caad: Claripy now pins Z3 5.0.0.0 on every platform, so there is no platform-specific solver version. I also tracked down the native ecosystem OOM rather than accepting the discrepancy: Z3 5 retains substantial native memory when Claripy creates many short-lived solvers, while Claripy’s existing per-thread solver-reuse mode avoids that accumulation. Solver reuse is now the default (and remains explicitly disableable via `REUSE_Z3_SOLVER=false`). On the prior `scanf_multi` reproducer, Z3 5 dropped from ~15.8 GB / 12 s to 282 MB / 1.3 s; the entire scanf file peaks at 288 MB. The full Claripy suite passes on Z3 5 (334 passed, 1 skipped), and the targeted angr scanf plus symbolic-sqrt tests pass.

zardus:
Correction after the full ecosystem tail completed: the “if the tests pass” condition is not met for native Z3 5. Current angr master has two symbolic floating-point callable tests (`test_manyfloatsum_symbolic_i386` and `test_manyfloatsum_symbolic_x86_64`) that expose a severe native Z3 5 regression. In an isolated A/B run, Z3 4.13 completed both in 42 seconds with a 692 MB peak. Z3 5 exceeded 4 GB on the first test after 38 seconds without completing; the parallel ecosystem shard grew past the runner memory budget. The QF_FPBV-specialized solver path did not improve it. I reverted the universal pin in 809d8907, restoring Z3 4.13 natively and Z3 5 only on Emscripten. The direct Claripy suite is green (332 passed, 1 skipped), and the two native FP regressions pass under the restored pin. This keeps the platform discrepancy only because using Z3 5 everywhere currently fails the broader test suite in practice.

twizmwazin:
Can you figure out the root cause of the issue in z3 and determine if there is a reasonable solution?

zardus:
I reduced this to a standalone Z3 problem; Claripy and angr are not involved in the failure. The reproducer is 19 FP64 variables constrained to be greater than `1.0`, a sequential RNE sum, and `sum == 27.7`:

```python
xs = [FP(f"x{i}", Float64()) for i in range(19)]
s = Solver()
s.add(*(fpGT(x, FPVal(1.0, Float64())) for x in xs))
total = functools.reduce(lambda a, b: fpAdd(RNE(), a, b), xs)
s.add(total == FPVal(27.7, Float64()))
s.check()
```

There are two upstream regressions:

1. The initial runtime regression first shipped in 4.13.1. Source-bisecting 4.13.0..4.13.1 identifies [Z3 `d66609ea`](https://github.com/Z3Prover/z3/commit/d66609ea14e8d329bb1ab730c93c98f3f67250be) (`fix #7389`). It added a reverse-constant `bvule` rewrite; the lower-bound comparisons emitted by FP bit-blasting hit that rule. Its parent solves the standalone case in ~12s; the introducing commit does not finish in 30s. The published 4.13.0 wheel solves in 8.6s/~519 MiB, while 4.13.1 through 4.13.4 all time out in the same search with substantially more restarts/subsumption work.

2. The large memory regression first shipped in 4.14.0. [Z3 `f41134d1`](https://github.com/Z3Prover/z3/commit/f41134d1b6f7462df23eb572addcc4cd46e2bd7b) refactored Boolean equality rewriting and makes this FP formula recurse to a stack overflow. [Z3 `6ea68310`](https://github.com/Z3Prover/z3/commit/6ea68310c971027b3bd788aa4ff6a6e0e195c398) fixes that stack overflow by deferring `not(eq(...))` back through the generic rewriter, but on this formula that becomes the rewrite/memory explosion seen in 4.14 and 5. At the introducing commit, a 10s run reaches ~3.7 GiB and creates only one SAT variable. The 4.14.0 wheel likewise spends the timeout in ~9M rewrite steps and exceeded 13 GiB by 50s; 4.13.4 reaches the SAT search at ~0.5 GiB.

I tested the available Claripy-side escape hatches:

- `SolverFor("QF_FPBV")`
- disabling `rewriter.bv_le2extract`, both globally and on the `qffp` tactic
- alternate SAT restart/branching settings (Luby, geometric, slow restart, CHB)
- solver reuse (which fixes a separate many-short-lived-solvers retention problem)

None rescues the case. The `bv_le2extract` option controls both the older useful rewrite and the new problematic direction, and there is no public option for the Boolean `not(eq)` rewrite.

So the reasonable solution that does not modify Z3 is the current split pin: exact 4.13.0 on native platforms, where the ecosystem FP tests complete, and Z3 5 on Emscripten, where Z3 4 is not WASM-host-compatible. A universal Z3 5 pin would require an upstream Z3 fix or a narrowly maintained Z3 patch; there is not a reliable Claripy-only parameter workaround. I have kept the minimized reproducer and source-bisect data if we want to turn this into an upstream Z3 report separately.

twizmwazin:
Explain what fixes are needed to address the regressions. Create a patch verifying this and test claripy/angr against it.

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

Follow-up with an actual Z3 source patch and integration results. One correction to my earlier reduction: the reduced z3py example used ordinary SMT equality, while Claripy emits `fp.eq`. I therefore extracted the exact SMT2 generated by the real i386 angr test and used that for this verification.

The verification patch makes four Z3-side changes:

1. Restrict the reverse-constant `bvule` rewrite added by [`d66609ea`](https://github.com/Z3Prover/z3/commit/d66609ea14e8d329bb1ab730c93c98f3f67250be) to the sign-extension shape for which it was introduced. Applying it to arbitrary right-hand sides is what sends this FP bit-blast into the bad SAT search.
2. In the one-bit `ite` equality rewrite changed by [`f41134d1`](https://github.com/Z3Prover/z3/commit/f41134d1b6f7462df23eb572addcc4cd46e2bd7b), construct the branch equalities without recursively invoking the Boolean rewriter. The enclosing `BR_REWRITE2` still performs the bounded follow-up rewrite.
3. Disable the expansive optional ITE rules by default. They reduce individual terms locally but amplify the post-bit-blast formula globally in this case.
4. Rewrite `fp.eq(x, literal)` to ordinary equality when the literal is non-NaN and nonzero. In that domain the relations are equivalent, and this avoids redundant NaN/zero/sign guards in the FP-to-BV lowering. Zero and NaN remain on the original path.

<details>
<summary>Verification patch against Z3 5.0.0 (8e3402b2)</summary>

```diff
diff --git a/src/ast/rewriter/bv_rewriter.cpp b/src/ast/rewriter/bv_rewriter.cpp
index b2ae76dd7..7305862b5 100644
--- a/src/ast/rewriter/bv_rewriter.cpp
+++ b/src/ast/rewriter/bv_rewriter.cpp
@@ -611,6 +611,33 @@ br_status bv_rewriter::mk_leq_core(bool is_signed, expr * a, expr * b, expr_ref
                     return i;
             return UINT_MAX;
         };
+
+        // Restrict the reverse rule to the sign-extension shape it was added
+        // for. Applying it to arbitrary terms with constant lower bounds can
+        // make the result much harder to solve.
+        auto is_sign_extended_concat = [&](expr* e) {
+            if (!m_util.is_concat(e))
+                return false;
+            app* c = to_app(e);
+            unsigned num_args = c->get_num_args();
+            if (num_args < 2)
+                return false;
+            expr* sign = c->get_arg(0);
+            unsigned i = 1;
+            while (i < num_args && c->get_arg(i) == sign)
+                ++i;
+            if (i == num_args)
+                return false;
+            expr_ref base(m);
+            base = i + 1 == num_args ? c->get_arg(i) : m_util.mk_concat(num_args - i, c->get_args() + i);
+            unsigned lo, hi;
+            expr* arg;
+            return m_util.get_bv_size(sign) == 1 &&
+                   m_util.is_extract(sign, lo, hi, arg) &&
+                   lo == hi &&
+                   hi + 1 == m_util.get_bv_size(base) &&
+                   arg == base;
+        };
         
         unsigned lnz = last_non_zero(b);
 
@@ -638,7 +665,7 @@ br_status bv_rewriter::mk_leq_core(bool is_signed, expr * a, expr * b, expr_ref
             result = m.mk_true();
             return BR_DONE;
         }
-        else if (lnz < bv_sz - 1 && m_le2extract) {
+        else if (lnz < bv_sz - 1 && m_le2extract && is_sign_extended_concat(b)) {
             // use the equivalence to simplify:
             // #x000f <=_u b <=> b[sz-1:lnz+1] != 0 or #xf <= b[lnz:0])
 
@@ -2532,8 +2559,8 @@ br_status bv_rewriter::mk_bit2bool(expr * lhs, expr * rhs, expr_ref & result) {
     expr* a = nullptr, *b = nullptr, *c = nullptr;
     if (m.is_ite(lhs, a, b, c)) {
         bool_rewriter rw(m);
-        expr_ref e1(rw.mk_eq_rw(b, rhs), m);
-        expr_ref e2(rw.mk_eq_rw(c, rhs), m);
+        expr_ref e1(rw.mk_eq_plain(b, rhs), m);
+        expr_ref e2(rw.mk_eq_plain(c, rhs), m);
         result = rw.mk_ite(a, e1, e2);
         return BR_REWRITE2;
     }
diff --git a/src/ast/rewriter/fpa_rewriter.cpp b/src/ast/rewriter/fpa_rewriter.cpp
index c2c888885..24f8b6a2f 100644
--- a/src/ast/rewriter/fpa_rewriter.cpp
+++ b/src/ast/rewriter/fpa_rewriter.cpp
@@ -473,6 +473,11 @@ br_status fpa_rewriter::mk_float_eq(expr * arg1, expr * arg2, expr_ref & result)
         result = (m_fm.eq(v1, v2)) ? m().mk_true() : m().mk_false();
         return BR_DONE;
     }
+    if ((m_util.is_numeral(arg1, v1) && !m_fm.is_nan(v1) && !m_fm.is_zero(v1)) ||
+        (m_util.is_numeral(arg2, v2) && !m_fm.is_nan(v2) && !m_fm.is_zero(v2))) {
+        result = m().mk_eq(arg1, arg2);
+        return BR_REWRITE1;
+    }
 
     return BR_FAILED;
 }
diff --git a/src/params/bool_rewriter_params.pyg b/src/params/bool_rewriter_params.pyg
index 87578470e..742e36f25 100644
--- a/src/params/bool_rewriter_params.pyg
+++ b/src/params/bool_rewriter_params.pyg
@@ -1,7 +1,7 @@
 def_module_params(module_name='rewriter',
                   class_name='bool_rewriter_params',
                   export=True,
-                  params=(("ite_extra_rules", BOOL, True, "extra ite simplifications, these additional simplifications may reduce size locally but increase globally"),
+                  params=(("ite_extra_rules", BOOL, False, "extra ite simplifications, these additional simplifications may reduce size locally but increase globally"),
                           ("flat", BOOL, True, "create nary applications for +,*,bvadd,bvmul,bvand,bvor,bvxor"),
                           ("flat_and_or", BOOL, True, "create nary applications for and,or"),
                           ("sort_disjunctions", BOOL, True, "sort subterms in disjunctions"),
```

Patch SHA-256: `9fc6f8a963608b93926da7cfe338476cb7eef1538ab699c877ed6d8a850354bb`

</details>

Validation used a Release+LTO `z3-solver==5.0.0.0` wheel built from that tree:

- Wheel SHA-256: `7b9d825ea39e35e8db28ce01abcf0ef0a05b2939ad042df548e845e680258323`.
- Exact extracted i386 SMT2: 3,820 bytes, 20 constraints, SHA-256 `d50eb63477beb0ed1e8d39fbf7f238a102362d6c32b0e0989fdd0e0747ac5103`; `sat` in 37.25 s with 305,844 KiB peak RSS.
- Full Claripy suite at `809d8907`: **332 passed, 1 skipped in 7.93 s**.
- The real angr regressions, using the same wheel:

  ```text
  tests/factory/test_callable.py::TestCallable::test_manyfloatsum_symbolic_i386
  tests/factory/test_callable.py::TestCallable::test_manyfloatsum_symbolic_x86_64
  2 passed in 139.94s
  ```

  Peak RSS was 684,924 KiB with no swap. Stock Z3 5 had exceeded 4 GiB without completing the first test.
- FP equivalence controls cover finite values, infinities, signed zero, and NaN. Patched and unpatched Z3 return the same six expected results; the optimization deliberately excludes zero and NaN. Patched `fp.eq` and ordinary equality also produce byte-identical output after `simplify/fpa2bv/simplify`.
- The original Z3 #7389 sign-extension case still simplifies correctly, and nested one-bit ITE equality controls for both `#b0` and `#b1` remain equivalent.
- `git diff --check` passes.

This is a verification/prototype patch, not something I would vendor into Claripy or submit upstream unchanged. In particular, changing the global `ite_extra_rules` default needs Z3's full regression suite and representative non-FP performance testing; the sign-extension recognizer is sound but intentionally conservative.

There is also a remaining runtime-only gap: setting `rewriter.bv_sort_ac=false` makes the same angr pair pass in 40.22 s at 588,328 KiB, but globally reverting Z3 5's new default could regress unrelated BV workloads such as [Z3 #7700](https://github.com/Z3Prover/z3/issues/7700). A QF_FP-local experiment still took 66.67 s for the exact `fp.eq` formula through the generic `Solver` API and did not reproduce that integration speedup, so I did not include a global rollback in the patch.

The result verifies that Z3-side changes can restore bounded memory and completion for the actual Claripy/angr workload. Until equivalent fixes are reviewed and released by Z3, the current PR's split pin remains the reasonable safe choice: Z3 4.13.0.0 natively and Z3 5.0.0.0 only where WASM host compatibility requires it. No further Claripy source change is needed for this concern.

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

Validation record for head `c871c135bf3abcdea4fd77a8263393707b009b16`, based on `master` `6ff4486278af191304ae8188ed6faa643cf43087` (master is at `0a2d0c6f3b619f358c9ee2f10e97b26a6290c416`; the branch is three commits behind it and `git merge-tree` reports no conflict).

This replaces the record written for `8739713e1681510aa518bfab72beecdfa3e2529d`, which validated the platform split. The split is gone; the branch now carries one unmarked `z3-solver==5.1.0.0`.

### What ran, and what did not

**No claripy test was run against this head outside CI, and none could be.** The workspace this branch was prepared in has no local claripy build and permits no virtual environment. What did run locally: `ruff check` and `ruff format --check` (ruff 0.16.5, repo config) clean on both changed Python files, `uv lock --check` exit 0, `pyproject.toml` parses. Everything else is CI on this head and the measurements linked below, which predate it.

### Platform coverage, re-derived 2026-09-08

All 52 `z3-solver` releases on PyPI, per-release wheel platform tags. Exactly one release carries `pyemscripten_2026_0_wasm32`: **5.1.0.0**, which is also the newest. Its other tags are `manylinux_2_27_x86_64`, `manylinux_2_38_aarch64`, `manylinux_2_38_riscv64`, `macosx_13_0_arm64`, `macosx_13_0_x86_64`, `win32`, `win_amd64`, `win_arm64`, plus an sdist.

5.0.0.0 carries that same set minus the Emscripten wheel. 4.16.0.0 is not the same set: it has no `manylinux_2_38_riscv64`, and its macOS wheels are `macosx_15_0` rather than `macosx_13_0`.

This raises the platform floor, and not only on glibc:

- Linux x86_64: `manylinux2014` (glibc 2.17) at 4.13.0.0 becomes `manylinux_2_27`.
- Linux aarch64: `manylinux2014` becomes `manylinux_2_38`, which excludes Ubuntu 22.04 (glibc 2.35) and Debian 12 (2.36).
- macOS: `macosx_11_0` becomes `macosx_13_0`.

Below those, pip falls back to building the sdist. That is upstream's packaging decision rather than this change's, and angr master already pins 5.1.0.0, so the ecosystem has taken the same floor.

### The FP sign ABI boundary, measured

`Z3_fpa_get_numeral_sign.argtypes`, read out of `z3/z3core.py` in one wheel from each of the 18 releases from 4.13.0.0 to 5.1.0.0:

| `POINTER(c_int)` | `POINTER(c_bool)` |
| --- | --- |
| 4.13.0.0, 4.13.1.0, 4.13.2.0, 4.13.3.0, 4.13.4.0, 4.14.0.0, 4.14.1.0, 4.15.0.0, 4.15.1.0, 4.15.3.0, 4.15.4.0 | 4.15.5.0, 4.15.6.0, 4.15.7.0, 4.15.8.0, 4.16.0.0, 5.0.0.0, 5.1.0.0 |

There is no 4.15.2.0 on PyPI, and 4.15.5.0, 4.15.6.0 and 4.15.7.0 are yanked, so 4.15.8.0 is the earliest installable release with the new signature. The change first shipped at 4.15.5.0, not at the 5.0.0.0 boundary an earlier revision of this branch assumed. Under an exact `==5.1.0.0` pin the runtime version test is dead code either way, so this head passes `ctypes.c_bool` directly rather than selecting between the two.

`Z3_fpa_get_numeral_sign` is reached only on the `FPVal` branch of `_abstract_fp_val` and `_abstract_fp_encoded_val`, so NaN, the infinities and the zeros never touch it. Nothing in `tests/test_fp.py` evaluated a finite symbolic FP back to a Python float before this branch — `test_nan` lands on `"NaN"`, and `test_negative_zero`, `test_fp_ops` and `test_fp_precision_loss` all evaluate BVs. `test_fp_model_value` is the regression for this change.

### The parameter

`rewriter.ite_extra_rules` measurements are at https://github.com/angr/claripy/pull/737#issuecomment-5561423603 (2026-09-06), on stock `z3-solver==5.1.0.0`, `libz3.so.5.1` sha256 `da99bbf1032d8f90698897c0408ade323440bdfda3304a7d1a00cc379d2acfc1` byte-identical between arms:

| arm | flag reads | test | outcome | `ulimit -v` | peak RSS |
| --- | --- | --- | --- | ---: | ---: |
| flag off + `c_bool` | `false` | `test_manyfloatsum_symbolic_i386` | pass | 24 GiB | 621,508 KiB |
| flag off + `c_bool` | `false` | `test_manyfloatsum_symbolic_x86_64` | pass | 24 GiB | 622,540 KiB |
| `c_bool` only | `true` | i386 | `solver unknown: out of memory` | 4 GiB | cap-bound |
| `c_bool` only | `true` | x86-64 | `solver unknown: out of memory` | 4 GiB | cap-bound |

The two arms did not share an address-space limit, so those columns are not matched, and neither limit can produce the result: the passing arm peaks at 2.5% of the ceiling it was given, and the same flag-less arm fails identically at 24 GiB. **The failing rows have no peak-RSS number and must not be given one.** Their high-water marks are set by the `ulimit -v`, not by z3 — the figure "about 4 GiB" was published on this pull request as a property of z3 and then retracted for exactly that reason: https://github.com/angr/claripy/pull/737#issuecomment-5465178840

That flag-off arm is the configuration this head is in, so the head is the measured arm rather than an extrapolation from it — but the measurement predates the head and was not re-run on it.

Two limits on the parameter, both stated when it was published. `z3.set_param` is process-global, so it reaches every other z3 user in the same interpreter; the precedent is `rewriter.hi_fp_unspecified`, set the same way since 4.8.7.0. And what it costs on non-FP workloads is not measured — claripy's own suite was 331 passed and 1 skipped with and without it, which is a weak control, because that pair also differed by the ABI change.

### What CI here can and cannot tell you

claripy's own suite runs in the `test` job: `uv sync --dev` and `uv run pytest`, three OSes by Python 3.12/3.13/3.14. That is the check to read for this change.

The `ecosystem` job cannot settle the FP question any more. angr/angr#6550 merged 2026-09-03 and removed the Python `claripy` dependency from angr, so the angr shards no longer import this package; `test_manyfloatsum_symbolic_*` now run against the Rust solver. `angr/ci-settings` `46dc618dd` (2026-09-07) has since removed `angr/claripy` from `ci-image/conf/repo-list.txt` and dropped `venv_install ./src/claripy` from `ci-image/scripts/ga-build.sh`; that lands when the `angr/ci` image is republished, and it will change what that job does here.

### Prior records

- Root cause and the four-change z3 prototype: https://github.com/angr/claripy/pull/737#issuecomment-5065397505
- Two numbers retracted: https://github.com/angr/claripy/pull/737#issuecomment-5465178840
- The measurement this head acts on: https://github.com/angr/claripy/pull/737#issuecomment-5561423603

twizmwazin:
Not merging unless it can use a single z3 version across all platforms

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

Agreed. No currently released Z3 version satisfies both requirements: 4.13.0.0 and 5.0.0.0 have no Emscripten wheel, while 5.1.0.0 has the full platform wheel set but stock and narrowly patched builds each fail both load-bearing native angr FP tests twice with Z3 `out of memory` at about 4 GiB RSS.

I am parking this PR as draft until upstream Z3 releases one version that is safe natively and available for Emscripten. The canonical exact-head record and released-version matrix are here: https://github.com/angr/claripy/pull/737#issuecomment-5427519925

---

**Superseded, 2026-09-08.** "About 4 GiB RSS" was my own `ulimit -v`, not z3's behaviour — retracted in https://github.com/angr/claripy/pull/737#issuecomment-5465178840 . And 5.1.0.0 is usable natively after all: turning off `rewriter.ite_extra_rules` makes both tests pass at about 621 MB, measured in https://github.com/angr/claripy/pull/737#issuecomment-5561423603 . The PR is no longer parked and no longer draft; it now carries one unmarked `z3-solver==5.1.0.0`.

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

Full dependency and signed-FP model state on Emscripten, before and after this change.

**Before** — the native-only pin and old ABI representation are unconditional:

<details><summary>Claripy baseline</summary>

```text
sys_platform == emscripten
selected dependency: z3-solver==4.13.0.0
Z3_fpa_get_numeral_sign output storage: ctypes.c_int
Pyodide-compatible Z3 wheel selected: no
```

</details>

**After** — Emscripten selects Z3 5 and its bool output ABI:

<details><summary>with this change</summary>

```text
sys_platform == emscripten
selected dependency: z3-solver>=5.0.0.0
Z3_fpa_get_numeral_sign output storage: ctypes.c_bool
Solver.eval(FPS == -2.5): (-2.5,)
Solver.eval(raw bits): (0xc0200000,)
```

</details>

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

Retracting two numbers I published on this pull request.

**"About 4 GiB RSS" was my own cap, not Z3's behaviour.** In the description, in the validation record and again in my comment of 2026-08-28, I wrote that Z3 5.1 fails the two native `manyfloatsum` tests with `out of memory` "at about 4 GiB RSS". Every one of those runs was under a `ulimit -v 6291456` — a 6 GiB address-space limit I set myself. Z3 reports `out of memory` when an allocation is refused, so the figure that gets recorded is the high-water mark just before the first request large enough to cross that limit. With a 6 GiB cap that lands near 4 GiB. Move the cap and it moves.

The tell is that the peaks are all the same number. Six failing trials under that cap — Z3 5.0.0.0 and 5.1.0.0, Claripy with and without the ABI fix, i386 and x86-64 — peaked between 4,013,016 KiB and 4,014,392 KiB. A 1.3 MiB spread across two solver versions, two Claripy variants and two targets describes the ceiling they shared, not the solver.

Move the ceiling and the failure point moves with it. Stock `z3-solver==5.1.0.0`, unmodified Claripy, `test_manyfloatsum_symbolic_i386`, one run per row, every run ending in the same `ClaripyZ3Error: solver unknown: out of memory`:

| `ulimit -v` | peak RSS | time to failure |
| --- | ---: | ---: |
| 6,291,456 KiB (6 GiB) | 4,013,436 KiB | 126 s |
| 10,485,760 KiB (10 GiB) | 7,805,332 KiB | 289 s |
| 12,582,912 KiB (12 GiB) | 7,807,424 KiB | 290 s |

It steps rather than scales: 6 GiB to 10 GiB nearly doubles the peak, 10 GiB to 12 GiB moves it by 0.03% because both are refused at the same allocation. So "about 4 GiB" is not a threshold in Z3 5.1 and should not be read as one. The exhaustion is real; where it would stop if left unbounded I did not measure, and I am not going to guess a second time.

**`rewriter.bv_sort_ac=false` does not rescue the pair on 5.1.0.0.** On 2026-07-24 I wrote that setting it makes the same pair pass in 40.22 s at 588,328 KiB. That was measured against a locally patched Z3 5.0.0.0 build. On the stock 5.1.0.0 wheel it does not work at all — the flag demonstrably takes effect (`get_param` reads `false` and `simplify(b + a)` stops reordering, through Claripy's own backend context), and both tests still fail with `solver unknown: out of memory`:

| test | `rewriter.bv_sort_ac` | result | time | peak RSS |
| --- | --- | --- | ---: | ---: |
| `test_manyfloatsum_symbolic_i386` | default | fail | 161 s | 4,016,756 KiB |
| `test_manyfloatsum_symbolic_i386` | `false` | fail | 103 s | 4,093,844 KiB |
| `test_manyfloatsum_symbolic_x86_64` | default | fail | 156 s | 4,014,188 KiB |
| `test_manyfloatsum_symbolic_x86_64` | `false` | fail | 79 s | 4,093,592 KiB |

all at `ulimit -v 6291456`; at 12 GiB the i386 case with the flag set reaches 11,677,964 KiB and fails identically. So there is no 5.1.0.0 figure to correct 40.22 s to — the sentence simply does not carry to the released version, and it should not be cited as a property of it.

Both retractions cut against the conclusion I drew from them, which was that no released Z3 is usable natively. I will follow this with the measurement that replaces it.

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

The measurement I promised in https://github.com/angr/claripy/pull/737#issuecomment-5465178840.

### The number

Stock `z3-solver==5.1.0.0`, unmodified Claripy, both load-bearing tests, one run at a time under `ulimit -v 25165824` — a 24 GiB address-space ceiling:

| test | peak RSS | VmPeak | outcome |
| --- | ---: | ---: | --- |
| `test_manyfloatsum_symbolic_i386` | 23,862,468 KiB | 24,436,632 KiB | `solver unknown: out of memory` |
| `test_manyfloatsum_symbolic_i386` (repeat) | 23,854,064 KiB | 24,436,644 KiB | same |
| `test_manyfloatsum_symbolic_i386` (second harness) | 23,861,092 KiB | 24,436,632 KiB | same |
| `test_manyfloatsum_symbolic_x86_64` | 23,853,076 KiB | 24,436,624 KiB | same |
| `test_manyfloatsum_symbolic_x86_64` (second harness) | 23,860,740 KiB | 24,436,612 KiB | same |

Eight runs have now ended this way at this ceiling — the five above plus three on the ABI-only arm. Their VmPeaks span 136 KiB, and every one stops about 712 MiB short of the 25,165,824 KiB limit. That is an allocation refused by `RLIMIT_AS`, not a solver reaching its own high-water mark.

**So the replacement for "about 4 GiB" is at least 22.7 GiB resident and 23.3 GiB of address space — and I owe this figure the same warning I gave the last one. It also stops short of a ceiling I chose.** "About 4 GiB" was an artifact of a 6 GiB `ulimit -v`; 23.3 GiB is an artifact of a 24 GiB one. I am not calling it unbounded, because I stopped at 24 GiB and did not measure past it.

These ran on 2026-09-02 and 2026-09-03, strictly sequentially — 35 measured cells, checked pairwise from the start and end timestamps each one records, zero overlapping pairs — on a machine that was busy throughout. That cannot move the VmPeak column: `RLIMIT_AS` is per-process address-space accounting, so contention has no bearing on it.

### What the number is of

Python Claripy at `6ff4486278af191304ae8188ed6faa643cf43087`, z3-solver **5.1.0.0**, and the nineteen-symbolic-argument form of these tests, which is what angr carried until [angr#7043](https://github.com/angr/angr/pull/7043) cut it to ten on 2026-08-30. Nobody has measured the ten-argument form at 5.1.0.0, including me.

### One line makes them pass

On the same stock wheel, `z3.set_param("rewriter.ite_extra_rules", "false")` plus the `Z3_fpa_get_numeral_sign` `c_int`-to-`c_bool` change this branch already carries:

| arm | flag reads | test | outcome | peak RSS |
| --- | --- | --- | --- | ---: |
| flag + ABI | `false` | i386 | pass | 621,508 KiB |
| flag + ABI | `false` | x86-64 | pass | 622,540 KiB |
| ABI only | `true` | i386 | `solver unknown: out of memory` | 4,011,564 KiB |
| ABI only | `true` | x86-64 | `solver unknown: out of memory` | 4,011,492 KiB |

Re-run on 2026-09-06, each arm proving from inside the pytest process which Claripy it loaded and what the parameter read. The two arms also had different address-space limits — 24 GiB for the passing rows, 4 GiB for the failing ones — so they are not matched on that; neither can produce the result, because the passing arm's peak RSS is 2.5% of the limit it was given, and three of the eight failures above are this same flag-less arm at 24 GiB. `libz3.so.5.1` is byte-identical between the arms, sha256 `da99bbf1032d8f90698897c0408ade323440bdfda3304a7d1a00cc379d2acfc1`. No Z3 was rebuilt or patched.

Neither change passes alone: without the ABI fix the tests reach floating-point model extraction and die with `ctypes.ArgumentError` at `backend_z3.py:757`.

### Corrections

- [2026-07-22](https://github.com/angr/claripy/pull/737#issuecomment-5052593249): "there is not a reliable Claripy-only parameter workaround." There is a Claripy-only parameter that removes the memory blow-up, and I never tried it. Whether it is safe across other workloads I still have not measured.
- [2026-07-24](https://github.com/angr/claripy/pull/737#issuecomment-5065397505): I presented disabling these rules as one change in a four-change Z3 C++ patch, flipping the compiled-in default. It is a public runtime parameter, settable from Python against the released wheel.
- [2026-08-28](https://github.com/angr/claripy/pull/737#issuecomment-5451458846) still says "at about 4 GiB RSS". Retracted on 2026-08-29; the figure to use is the one above.

### Where this belongs now

[angr#6550](https://github.com/angr/angr/pull/6550) merged on 2026-09-03. The Python `claripy` package is gone from angr, replaced by the Rust `clarirs` crates; `angr.claripy` is stub-only and angr no longer depends on `claripy`. So I ran these tests on that stack, at nineteen symbolic arguments, from a tests tree differing from the checkout only in its symbolic-argument count:

| stack | symbolic args | test | outcome | pytest time | peak RSS |
| --- | ---: | --- | --- | ---: | ---: |
| Rust `clarirs` | 19 | i386 | pass | 14.61 s | 457,572 KiB |
| Rust `clarirs` | 19 | x86-64 | pass | 13.60 s | 458,056 KiB |
| Python Claripy | 19 | i386 | `solver unknown: out of memory` | 156.93 s | cap-bound |
| Python Claripy | 19 | x86-64 | `solver unknown: out of memory` | 152.96 s | cap-bound |

Every row is z3-solver **5.0.0.0** — a different version from the rest of this comment — under a 6 GiB ceiling. The Rust rows were run on 2026-09-06. The Python rows are from 2026-08-29 and shared the machine with another job, taking 81% and 79% of a CPU, so their wall times are inflated; what they establish is that the run did not complete, not how long it took. The Rust peaks are genuine high-water marks at about 7% of that ceiling; the Python runs are stopped by it, which is why I quote no memory ratio. The two checkouts are 46 commits apart, `a9ca247a5` against `854269112`, so the frontend is not the only difference between them.

I said no released Z3 was usable natively. That is not a property of released Z3 on its own: the same released 5.0.0.0, on the same workload, either exhausts memory or completes in fifteen seconds depending on which frontend drives it. Why it does I have not established. The blow-up did not follow claripy into angr.

What I have not measured is the Rust stack on 5.1.0.0, which is the version Pyodide needs; the built extension links `libz3.so.5.0`, so that needs a rebuild rather than a wheel swap.

### What this does not establish

Two named tests, on Linux, on one machine. `z3.set_param` is process-global, so it reaches any other Z3 user in the same interpreter — the precedent is the `rewriter.hi_fp_unspecified` line Claripy has set the same way since 4.8.7.0. I have not measured what the flag costs on non-FP workloads; Claripy's own suite is unchanged between stock and the flag-plus-ABI arm, 331 passed and 1 skipped in both, but that pair differs by the ABI change too. This branch does not carry the flag: its head is `8739713e1681510aa518bfab72beecdfa3e2529d`, with the ABI fix and the split pin. I am reporting a measurement, not proposing a merge.

`angr/claripy` still publishes — 9.3.4 went to PyPI on 2026-09-02 with `z3-solver==4.13.0.0` — so a change here still reaches standalone users, but it can no longer change what `pip install angr` gets. Your condition is met by a single unmarked `z3-solver` pin rather than the platform split this branch carries; the `set_param` line is a separate thing, and it is what makes the tests pass at a 5.1.0.0 pin.

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

One version, every platform: `z3-solver==5.1.0.0`, unmarked. Pushed as `c871c135`, out of draft. The platform split is gone.

5.1.0.0 is the only `z3-solver` release with an Emscripten wheel — checked across all 52 releases on 2026-09-08 — so it is the single version that covers Pyodide and native at once. It is also what angr pins, so claripy and angr are co-installable again; at 4.13.0.0 against 5.1.0.0 they were not.

What unblocks the native side is z3's `rewriter.ite_extra_rules`. With it off, `test_manyfloatsum_symbolic_i386` and `_x86_64` pass at about 621 MB on the stock 5.1.0.0 wheel; with it on they exhaust memory. Measurements: https://github.com/angr/claripy/pull/737#issuecomment-5561423603 — and to be clear about what that is, it is a mitigation, not the cause. The parameter defaults to on in both 4.13.0 and 5.1.0 and 4.13.0.0 passes with it on; the underlying regression is the one bisected further up this thread.

The branch sets it beside the `rewriter.hi_fp_unspecified` line claripy has set the same way since 4.8.7.0. It is process-global, and that is the part I would rather you looked at than took from me: claripy's own suite is unchanged with it set, but what it costs on non-FP workloads I have not measured.

Also in: `ctypes.c_bool` at both `Z3_fpa_get_numeral_sign` sites. That argtype went from `POINTER(c_int)` to `POINTER(c_bool)` in 4.15.5.0, so without the change reading a floating-point model value back through that call raises `ArgumentError`. No test read one back; there is one now.

One cost worth naming: 5.1.0.0 raises the wheel floor. Linux x86_64 goes from `manylinux2014` to `manylinux_2_27`, aarch64 to `manylinux_2_38` (which excludes Ubuntu 22.04 and Debian 12), macOS from 11.0 to 13.0. angr already took that floor when it pinned 5.1.0.0.

Validation record: https://github.com/angr/claripy/pull/737#issuecomment-5427519925

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

Closing this without merging.

`angr/claripy` is deprecated in favour of claripy inside `angr/angr`, so a change
made here no longer reaches the code that runs. Reopening or re-targeting this
pull request is not the right move either: the work belongs against claripy in
`angr/angr` if it is still wanted there.

For anyone who picks this up, what this change did and what was established
while making it:

- It replaced the platform-split z3 dependency with a single unmarked
  `z3-solver==5.1.0.0`, so one z3 version is used on every platform.
- The `Z3_fpa_get_numeral_sign` ctypes signature changes from `c_int` to
  `c_bool` at z3 **4.15.5.0**. Two claims made earlier in this branch's own
  history were wrong about that boundary and are withdrawn: it is not 4.16.0.0
  and it is not 5.0.0.0.
- `z3.set_param("rewriter.ite_extra_rules", "false")` is a **mitigation, not a
  fix for a z3 5 regression**. `bool_rewriter_params.pyg` is identical at tags
  `z3-4.13.0` and `z3-5.1.0`, with `ite_extra_rules` defaulting to `True` in
  both, so the behaviour it guards against is not new in z3 5.
- PyPI carries 52 `z3-solver` releases and exactly one Emscripten wheel
  (5.1.0.0), which is what made a single pinned version possible at all.

No further work is planned on this pull request.

PR #742, opened 2026-08-25
Answer singlevalued and multivalued when no backend can bound cardinality

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

### Problem

`singlevalued` and `multivalued` raise instead of answering for any expression no backend can bound. On claripy master:

```
$ python -c "import claripy; x = claripy.BVS('x', 32); print(claripy.RotateLeft(x, claripy.BVS('amount', 32)).singlevalued)"
claripy.errors.BackendError: All backends failed to cardinality
```

`claripy.fpToFP(x, claripy.FSORT_FLOAT)` does the same, and so does `multivalued` on both. angr reads these predicates in its memory guards — the three `singlevalued` tests in `angr/storage/memory_mixins/simple_interface_mixin.py`, at lines 35, 51 and 62 — so the raise escapes as a `BackendError` through `CFGFast` and loses a whole binary rather than one block.

### Root cause

Both predicates are answered through `cardinality`. `BackendZ3` does not implement it at all, and for a rotate by a symbolic amount `BackendConcrete` declines (symbolic) and `BackendVSA` declines (no representation), so `BackendAny._first_backend` runs out of backends. It returned `None` there, which is falsy and gave callers the conservative answer, until `22c0039f` added a terminating `raise BackendError(f"All backends failed to {what}")`. Every caller written against the old shape started getting an exception where it had been reading a boolean.

### Fix

`BackendAny.singlevalued` and `BackendAny.multivalued` catch `BackendError` and answer `False` and `True`. Failing to bound the cardinality is not evidence that the expression holds one value. Same two expressions, this head:

```
fpToFP(BVS('x', 32), FSORT_FLOAT).singlevalued -> False
fpToFP(BVS('x', 32), FSORT_FLOAT).multivalued  -> True
RotateLeft(BVS('x', 32), BVS('amount', 32)).singlevalued -> False
RotateLeft(BVS('x', 32), BVS('amount', 32)).multivalued  -> True
```

`cardinality` itself keeps raising: there is no safe number to invent, but there is a safe answer to "is this one value". The base `Backend.singlevalued` and `.multivalued` gained `-> bool` because pyright inferred the chain through the raising `Backend._cardinality` as `NoReturn`, and a `NoReturn` base cannot be overridden compatibly.

This changes the exception type angr surfaces, which the diff does not show. The three guards above raise `SimMemoryError` where they propagated `BackendError`; `angr/analyses/variable_recovery/variable_recovery.py` (185, 217) logs a warning and returns instead of raising; `angr/concretization_strategies/signed_add.py` (19, 24) stops swapping rather than raising. angr has exactly one `except BackendError` in the package and it is in none of these paths, so no companion change is needed.

### Testing

`tests/test_expression.py::test_singlevalued_when_no_backend_can_bound_cardinality` asserts that every backend raises `BackendError` from `cardinality` for both expressions, then that `singlevalued` is `False`, `multivalued` is `True`, and `cardinality` still raises. The rotate case carries no floating point at all, which is what pins the hole as `cardinality` rather than FP support. It fails on the baseline with `BackendError: All backends failed to cardinality` and passes here; `pytest tests` reports 332 passed, 1 skipped and 2 subtests passed.

No public binary reproduces this — carrying a data-dependent x87 stack operation is necessary but not sufficient, and all 15 candidates in `angr/binaries` ran under `CFGFast` without raising — so the regression is a synthetic AST, which pins the defect directly and needs no fixture.

Validation: https://github.com/angr/claripy/pull/742#issuecomment-5418527215

session: emo-corpus


zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

Validation record for head `f3cb740fe0afde53bcff1894c7e9e0af2f3eb5c9` against baseline `6ff44862`. Everything below was measured at `f0f35760` and re-checked at `f3cb740f`; the update at the end says what changed and why the earlier measurements still hold.

- Regression: `pytest tests/test_expression.py -k singlevalued_when_no_backend` — **observed** failing on baseline, not merely asserted. A restore of the fix silently did nothing (`git -C <repo> apply <relative-path>` resolves the path against the repo, not the caller's directory), so the suite ran with `backend_any.py` reverted and the new test present. It failed with exactly `claripy.errors.BackendError: All backends failed to cardinality`. Re-applied with an absolute path; green again.
- Focused: `pytest tests` — 332 passed, 1 skipped, 2 subtests passed. The 2 subtests are the FP and rotate cases.
- Workspace gate: `ANGR_FEATURE=vex-geti-index nix develop --command ./.agents/skills/angr-validate-workspace/scripts/run-all-tests.sh` — all selected suites passed. claripy 332 passed / 1 skipped / 2 subtests; angr Python 2508 passed / 46 skipped / 2 xfailed / 260 subtests (27:49); angr Rust 35 passed; pre-commit over all files, test-inputs and workspace checks passed; both worktrees clean afterwards. angr was adopted alongside claripy specifically to cover the downstream exposure, since angr reads these two predicates.
- Per-backend behaviour on `fpToFP(BVS("x", 32), FSORT_FLOAT)`: `BackendConcrete` declines (symbolic), `BackendZ3` declines (no `cardinality`), `BackendVSA` declines (no FP representation). `RotateLeft(BVS("y", 32), BVS("amount", 32))` behaves identically with no floating point involved.

Reproducer, no fixture required:

```
python -c "
import claripy
x = claripy.BVS('x', 32)
print(claripy.fpToFP(x, claripy.FSORT_FLOAT).singlevalued)
print(claripy.RotateLeft(x, claripy.BVS('amount', 32)).singlevalued)
"
```

On baseline both lines raise `BackendError: All backends failed to cardinality`; on this head both print `False`.

How often this is reachable through angr, stated as a bound rather than a rate:

| Corpus | Enumeration | Carrying a data-dependent x87 stack op | Reaching the guard |
| --- | --- | ---: | ---: |
| decbench | all 803 binaries, static disassembly | 0 | — |
| angr/binaries | all 820 ELF/PE under 8 MB, static disassembly | 15 | 0 |

- Instrument control: the same command finds 62 such instructions in a binary known to trigger the failure, so the zeros are measured rather than an absent tool. An earlier run of this scan was void because `objdump` was not on `PATH` and every failed invocation was laundered into a zero.
- All 15 candidates were run under `CFGFast` on an unpatched tree: 14 completed, 1 failed to load, none raised. Largest were `x86_64/windows/Project1.vmp.exe` at 709 s and `i386/windows/a71a3c3b...` recovering 29,062 functions in 397 s, which supersedes an earlier 600 s timeout on the latter.
- Carrying the instruction is necessary but not sufficient: the floating-point operand must also be unconstrained where it feeds the register index. That is why 15 candidates yielded nothing, and why a larger corpus sweep is not the missing ingredient.

Caveats:

- No public binary reproduces this. The regression is therefore a synthetic AST, which pins the defect directly and needs no fixture.
- Call-site hardening in angr's `SimpleInterfaceMixin` guards was written and measured to be exactly as effective as this change, then discarded as redundant: the same three guards read `.singlevalued`, so fixing the predicate covers all of them.
- The gate ran with angr adopted alongside claripy, to cover the downstream exposure, but that angr worktree sits at `84684e2a7`, three commits behind angr master at the time.



---

**Update for head `f3cb740f` (2026-08-28).** `Lint` and `Typecheck` were red at `f0f35760`; both are green now, along with the other 24 checks on this head. Three findings, none of them cosmetic in the same way:

- `Typecheck`, `backend_any.py` badness `0.0 -> 0.222` (2 errors / 90 lines, reproduced locally to the digit). `Backend.singlevalued` and `Backend.multivalued` are answered through `Backend._cardinality`, whose base implementation only raises, so pyright inferred the whole chain as `NoReturn`. Nothing can override a `NoReturn` method compatibly, since `Never` is the bottom type -- annotating the *overrides* cannot fix this, only the base can. `NoReturn` is also the wrong description of a method subclasses are meant to implement, so the base now declares the `bool` its callers already assume: `claripy.ast.Base.singlevalued` and `.multivalued` are the only consumers and are both annotated `-> bool` already, and no other `Backend` subclass overrides either method. Badness `0.222 -> 0.0`; `backend.py` itself joins the diff at an unchanged `10.00` lint and unchanged `0.0245` badness.
- `Lint`, `tests/test_expression.py` `9.98 -> 9.96`, from `W0640 cell-var-from-loop` on the new test's `assertRaises(..., lambda: expr.cardinality)`. The `# noqa: B023` silenced ruff and not pylint. The loop variable is now bound as a lambda default, which satisfies both, and the `noqa` is gone.
- Same file, `R0124 comparison-with-itself` on the pre-existing `assert not claripy.backends.z3.is_false(a == a)` in `test_true_false_cache`. This is a false positive: `a` is an AST, so `a == a` builds a symbolic equality expression rather than comparing a value with itself, and that expression being neither true nor false is precisely what the test asserts. Suppressed narrowly with a `pylint:disable` and a comment saying why, rather than rewriting the assertion. File now scores `10.00`, above its own baseline.

Re-validation at `f3cb740f`:

- Merge-base comparison, the same one CI runs, per changed file: lint `backend.py 10.00 -> 10.00`, `backend_any.py 10.00 -> 10.00`, `test_expression.py 9.98 -> 10.00`; badness `backend.py 0.0245 -> 0.0245`, `backend_any.py 0.0 -> 0.0`, `test_expression.py 1.966 -> 1.895`. The reproduction was checked against CI rather than merely passing: pylint on the unfixed head printed `9.96/10` with exactly `cell-var-from-loop` and `comparison-with-itself`, and pyright printed `0.2222` on `backend_any.py`, both matching the numbers the red jobs reported.
- `pytest tests` — 332 passed, 1 skipped, 2 subtests passed, unchanged from the previous head.
- The repository's pinned `pre-commit` over the three changed files — all hooks pass, nothing rewritten.
- The measurements above this update were **not** regenerated. The only production change in this commit is two return annotations, which are not executed as part of either method's body and cannot move a corpus result, a backend's decline, or the reproducer's output. The complete workspace gate was not re-run for this head either; the ten-shard ecosystem matrix on this head is green, which covers the downstream angr exposure the earlier gate was adopted for.
- `Decompiler Snapshot Testing` succeeded on this head and `angr/claripy_742` sits exactly on `dec-snapshots` master (`compare` reports `identical`), i.e. byte-identical corpus output -- as expected for an annotation-only production change.

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

Correction to my validation comment: "all selected suites passed" did not say what was left unselected. That shell had adopted `claripy` and `angr` only; the `archinfo`, `pypcode`, `pyvex`, `pysoot`, `cle` and `angr-management` suites did not run. The claripy and angr suites — the two this change can affect — did run, so the verdict stands, but the green covers those two rather than the workspace.

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

Addition to the compatibility note above: this changes the *exception type* angr surfaces in several places, which the description does not say.

`singlevalued`/`multivalued` raising `BackendError` today means `angr/storage/memory_mixins/simple_interface_mixin.py` (lines 35, 51, 62) propagates `BackendError` out of `_translate_addr`; with this change those guards see `False` and raise `SimMemoryError` instead. Verified with the change monkeypatched in, using `claripy.RotateLeft(BVS, BVS)` as the address expression.

Two other subsystems change behaviour the same way: `angr/analyses/variable_recovery/variable_recovery.py` (185, 217) goes from raising to logging a warning and returning, and `angr/concretization_strategies/signed_add.py` (19, 24) stops swapping rather than raising.

Nothing breaks — angr has exactly one `except BackendError` in the package and it is in none of these paths — so no companion PR is needed. Recording it so the behaviour change is on the record.

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

Full output of both predicates on the two expressions the regression covers, before and after this change. The reproducer needs no fixture:

```
python - <<'EOF'
import claripy
x = claripy.BVS("x", 32)
for label, expr in (
    ("fpToFP(BVS('x', 32), FSORT_FLOAT)", claripy.fpToFP(x, claripy.FSORT_FLOAT)),
    ("RotateLeft(BVS('x', 32), BVS('amount', 32))",
     claripy.RotateLeft(x, claripy.BVS("amount", 32))),
):
    for attr in ("singlevalued", "multivalued"):
        try:
            print(f"{label}.{attr} -> {getattr(expr, attr)}")
        except Exception as e:
            print(f"{label}.{attr} -> {type(e).__module__}.{type(e).__name__}: {e}")
EOF
```

**Before** — every backend declines `cardinality`, so both predicates raise instead of answering, and angr's memory guards propagate that out of `CFGFast` rather than rejecting one address:

<details><summary>claripy master (6ff44862)</summary>

```
fpToFP(BVS('x', 32), FSORT_FLOAT).singlevalued -> claripy.errors.BackendError: All backends failed to cardinality
fpToFP(BVS('x', 32), FSORT_FLOAT).multivalued -> claripy.errors.BackendError: All backends failed to cardinality
RotateLeft(BVS('x', 32), BVS('amount', 32)).singlevalued -> claripy.errors.BackendError: All backends failed to cardinality
RotateLeft(BVS('x', 32), BVS('amount', 32)).multivalued -> claripy.errors.BackendError: All backends failed to cardinality
```

</details>

**After** — failing to bound the cardinality answers the conservative way round; `cardinality` itself still raises, because there is no safe number to invent:

<details><summary>with this change (f3cb740f)</summary>

```
fpToFP(BVS('x', 32), FSORT_FLOAT).singlevalued -> False
fpToFP(BVS('x', 32), FSORT_FLOAT).multivalued -> True
RotateLeft(BVS('x', 32), BVS('amount', 32)).singlevalued -> False
RotateLeft(BVS('x', 32), BVS('amount', 32)).multivalued -> True
```

</details>

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

Taking the close as a rejection of the shape rather than of the problem, and not asking for it back. Three things for the record, then a question about which direction you would rather we take.

**The behaviour is still there.** claripy `6ff44862`, which is master today:

```python
import claripy
x = claripy.BVS("x", 32)
claripy.fpToFP(x, claripy.FSORT_FLOAT).singlevalued
claripy.RotateLeft(x, claripy.BVS("amount", 32)).multivalued
```

```
fpToFP(BVS('x',32), FSORT_FLOAT).singlevalued            -> claripy.errors.BackendError: All backends failed to cardinality
fpToFP(BVS('x',32), FSORT_FLOAT).multivalued             -> claripy.errors.BackendError: All backends failed to cardinality
RotateLeft(BVS('x',32), BVS('amount',32)).singlevalued   -> claripy.errors.BackendError: All backends failed to cardinality
RotateLeft(BVS('x',32), BVS('amount',32)).multivalued    -> claripy.errors.BackendError: All backends failed to cardinality
```

**Correcting this pull request's own description.** It said the exception "escapes as a `BackendError` through `CFGFast` and loses a whole binary". That is wrong as written: `SimpleInterfaceMixin` is not in `DefaultMemory`'s MRO, so the default path never reaches the guard. The escape is real but narrower — it goes through `FastMemory` and through the decompiler's `FasterMemory`. On angr `a9ca247a5`:

```
FastMemory.load(RotateLeft(BVS('x',32), BVS('amount',32))) -> claripy.errors.BackendError: All backends failed to cardinality
FastMemory.load(plain BVS('plain',32)  [control])          -> angr.errors.SimMemoryError: address not supported
```

The control line is the whole point: with an address VSA can bound, the guard raises the `SimMemoryError` it means to; with one it cannot, a raw claripy exception escapes angr's memory model.

**Why we are not re-filing this patch.** angr#6550 replaces the Python claripy with `angr/claripy` and `native/clarirs-*`. There is no `backend_any` in that surface, so a `try`/`except` there buys nothing past the migration.

So: which would you rather see? Harden the three `SimpleInterfaceMixin` guards in angr to treat an unanswerable cardinality as "not single-valued", which is angr-side and survives clarirs; or give `BackendZ3` the `_cardinality` its own base docstring asks for, so the over-approximation exists rather than being caught after the fact. We will build whichever you name, and neither if you would rather leave it.

session: sharpen

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

Withdrawing the question I left here on 2026-08-29. It no longer needs an answer.

That comment asked whether you would rather we harden the three `SimpleInterfaceMixin` guards in angr or give `BackendZ3` the `_cardinality` its base docstring asks for. The clarirs migration (angr/angr#6550, angr master `8b8d7a610`) landed five days later and settled it. Measured on angr `854269112`, three commits behind master and none of them in this area:

```
RotateLeft(BVS('x',32), BVS('amount',32)).singlevalued -> False
RotateLeft(BVS('x',32), BVS('amount',32)).multivalued  -> True
RotateLeft(BVS('x',32), BVS('amount',32)).cardinality  -> 4294967296
FastMemory().load(that address, size=4) -> angr.errors.SimMemoryError: address not supported
```

That is the answer this patch was trying to produce, and angr's memory guard now raises its own error instead of letting a claripy one out.

Standalone claripy still raises `BackendError: All backends failed to cardinality` for the same two expressions on master `0a2d0c6f`, but angr master has no `claripy` dependency and no `import claripy` left, so we are not proposing a change here. Nothing needed from you; leaving this closed.

session: sharpen

